PR #1, opened 2019-09-06
Add SshConnection class

tmds:
CC @aslicerh

### DIFF
--- a/src/Tmds.Ssh/SshConnection.cs
+++ b/src/Tmds.Ssh/SshConnection.cs
@@ -1,0 +1,40 @@
+﻿// This file is part of Tmds.Ssh which is released under LGPL-3.0.
+// See file LICENSE for full license details.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Tmds.Ssh
+{
+    public sealed class SshConnection : IDisposable
+    {
+        private readonly SshConnectionSettings _settings;
+        private readonly ILogger _logger;
+
+        public SshConnection(SshConnectionSettings settings, ILogger logger = null)
+        {
+            _settings = settings ?? throw new ArgumentNullException(nameof(settings));
+            _logger = logger;
+        }
+
+        public async Task ConnectAsync(CancellationToken cancellationToken)
+        {
+            // TODO: SshConnectionSettings needs an upper bound time for this method (e.g. SshConnectionSettings.ConnectTimeout)
+        }
+
+        // This method is for doing a clean shutdown which may involve sending some messages over the wire.
+        public async Task DisconnectAsync(CancellationToken cancellationToken)
+        {
+            // TODO: SshConnectionSettings needs an upper bound time for this method (e.g. SshConnectionSettings.DisconnectTimeout)
+
+            // In a finally block, this method calls Dispose.
+        }
+
+        // This method will just cut the connection.
+        public void Dispose()
+        {
+        }
+    }
+}
--- a/src/Tmds.Ssh/Tmds.Ssh.csproj
+++ b/src/Tmds.Ssh/Tmds.Ssh.csproj
@@ -1,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+   <FrameworkReference Include="Microsoft.AspNetCore.App" />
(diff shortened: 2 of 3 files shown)


src/Tmds.Ssh/SshConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+// See file LICENSE for full license details.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Tmds.Ssh
+{
+    public sealed class SshConnection : IDisposable
```
aslicerh: Just to confirm, we're only going to support ssh v2? No backwards 1.x support, right?

src/Tmds.Ssh/SshConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+        // This method is for doing a clean shutdown which may involve sending some messages over the wire.
+        public async Task DisconnectAsync(CancellationToken cancellationToken)
+        {
+            // TODO: SshConnectionSettings needs an upper bound time for this method (e.g. SshConnectionSettings.DisconnectTimeout)
+
+            // In a finally block, this method calls Dispose.
+        }
+
+        // This method will just cut the connection.
+        public void Dispose()
```
aslicerh: Just a thought, if dispose also tries to disconnect cleanly if still connected when called (maybe treat it as is the local connection quit due to a timeout if no other reason is given?), then one could abandon a connection object to the heap/GC and still possibly close cleanly. Kind of a pseudo RAII approach to handling connection objects.

I haven't really thought the effects through completely but it was just something that came to mind.

src/Tmds.Ssh/SshConnection.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Tmds.Ssh
+{
+    public sealed class SshConnection : IDisposable
+    {
+        private readonly SshConnectionSettings _settings;
```
aslicerh: It might be good to be able to retrieve a connection object's settings after it has been created. Maybe for classification purposes (who it is connected to, what port is it using, what crypto protocols does it use, etc.), or to possibly duplicate them after the fact (like 'restarting' a previously connected session).

So maybe a getter property here would be appropriate?

src/Tmds.Ssh/SshConnection.cs
tmds: Yes, only things which are 'recent'.

src/Tmds.Ssh/SshConnection.cs
tmds: The reason for not trying a clean disconnect is that the user has no cancellationtoken to indicate how long he wants to spend on that. We can have that on an explicit Disconnect method.
```
// All of this may take at most 500ms:
ct = new CancellationTokenSource(500).Token;
using var connection = new SshConnection(...);
await connection.ConnectAsync(ct);
await connection.DoStuffAsync(ct);
await connection.DisconnectAsync(ct); // try a clean disconnect
```
Probably this will change to `IAsyncDisposable`, which suffers from some keyword verbosity:
```
await using var connection = ...
```

src/Tmds.Ssh/SshConnection.cs
tmds: > It might be good to be able to retrieve a connection object's settings after it has been created.

I'm experimenting with a certan design here which separates these concerns:
- providing settings that control the connection behavior -> no getter for settings
- calling connect -> no getter to know if connectasync has been called

We can add them later if we want to. Removing them later is more difficult.

> Maybe for classification purposes (who it is connected to, what port is it using, what crypto protocols does it use, etc.), or to possibly duplicate them after the fact (like 'restarting' a previously connected session).

Yes, providing such info would be interesting. We should put them on a separate object, like `ConnectionInfo`.

src/Tmds.Ssh/Tmds.Ssh.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+   <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
tmds: fyi, this is to be able to use `System.IO.Pipelines` because it is not part of the .NET Core shared framework, but of the ASP.NET Core shared framework.

PR #2, opened 2019-09-12
Do some design

tmds:
This describes down how data will flow and SSH channels implementation hooks into the `SshConnection`

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,98 @@
+SSH is a multiplexed, packet-oriented protocol.
+
+This is different from SSL which is stream-oriented.
+And HTTP which is also stream-oriented.
+HTTP/2 is a stream-oriented, multiplexed protocol: it multiplexes/demultiplexes a number of streams into one.
+
+Because SSH is a packet-oriented, we can't use the [Pipe](https://docs.microsoft.com/en-us/dotnet/api/system.io.pipelines.pipe?view=dotnet-plat-ext-2.1) class after we've decrypted the packets.
+The `Pipe` class represents a stream buffer, and storing the packets into it would make us lose the packet boundaries.
+
+for buffering packets, we can make use of the [Channel](https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels.channel?view=dotnet-plat-ext-2.1) class.
+
+As a type to represent packets, we want something that works wel together with the `Pipe` related types, like [Memory](https://docs.microsoft.com/en-us/dotnet/api/system.memory-1?view=netstandard-2.1), [ReadOnlySequence<T>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.readonlysequence-1?view=netstandard-2.1), and [SequenceReader<T>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.sequencereader-1?view=netcore-3.0).
+
+We want to pool the packet type so it can be re-used by the `SshConnection`.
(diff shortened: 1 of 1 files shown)


tmds:
cc @aslicerh @omajid

DESIGN.md
```diff
@@ -0,0 +1,98 @@
...
+for buffering packets, we can make use of the [Channel](https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels.channel?view=dotnet-plat-ext-2.1) class.
+
+As a type to represent packets, we want something that works wel together with the `Pipe` related types, like [Memory](https://docs.microsoft.com/en-us/dotnet/api/system.memory-1?view=netstandard-2.1), [ReadOnlySequence<T>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.readonlysequence-1?view=netstandard-2.1), and [SequenceReader<T>](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.sequencereader-1?view=netcore-3.0).
+
+We want to pool the packet type so it can be re-used by the `SshConnection`.
+Also, if an instance was parsed, we may want to trim it and allow passing it to the user.
+
+So our packet should be some type of _mutable_ counterpart of `ReadOnlySequence`.
+```cs
+class Sequence
```
davidfowl: Should this implement `IBufferWriter<byte>`?

DESIGN.md
```diff
@@ -0,0 +1,98 @@
...
+    // Writer API
+    Memory<byte> AllocMemory(int sizeHint); // PipeWriter.GetMemory
+    Span<byte> AllocSpan(int sizeHint);     // PipeWriter.GetSpan
+    void Append(int);                       // PipeWriter.Advance (Adds data put into the Allocated Span/Memory)
+    void Append(Sequence);                  // Extracts data from other sequence and adds it to this sequence.
+    // Reader API
+    void Remove(int);                       // PipeReader.AdvanceTo (Removes data from the front of the Sequence)
+    ReadOnlySequence<byte> GetReadOnlySequence();     // Allows the Sequence to be read, e.g. by SequenceReader
+    ReadOnlySequence<byte> ExtractReadOnlySequence(); // Extracts data from the Sequence
+    // Pooling
```
davidfowl: These 2 APIs look strange. What's the difference?

DESIGN.md
davidfowl: Also see https://github.com/AArnott/Nerdbank.Streams/blob/master/src/Nerdbank.Streams/Sequence%601.cs

tmds: > Should this implement IBufferWriter<byte>?

It meets the contract.

> Also see https://github.com/AArnott/Nerdbank.Streams/blob/master/src/Nerdbank.Streams/Sequence%601.cs

This is similar to that type, but tied to `SequencePool` to provide pooling of `Memory`, `Segments`, and the `Sequences` themselves.

DESIGN.md
tmds: If you call the `ExtractReadOnlySequence`, the `Sequence` no longer owns the `Segments` and `Memory`. Next calls to `GetReadOnlySequence` will return an `Empty` `ReadOnlySequence<byte>`. The caller owns the `ReadOnlySequence<byte>` that was handed to him.

tmds:
This is meant to set some direction, merging as-is.

DESIGN.md
davidfowl: I’m not a fan of that subtlety

tmds: The alternative is to copy data which doesn't need to be copied.

PR #23, opened 2019-10-28
Figure out how to calculate shared secret using BouncyCastle

tmds:


tmds:
This takes a dependency on BouncyCastle to calculate the ecdh shared secret cross platform.

tmds:
cc @omajid @aslicerh

PR #68, opened 2020-04-22
ChannelContext: fix sendwindow implementation

tmds:


src/Tmds.Ssh/SshClient.ChannelContext.cs
```diff
@@ -365,25 +353,16 @@ public override async ValueTask SendChannelDataAsync(ReadOnlyMemory<byte> memory
...
-                    if (sendWindow > 0)
+                    try
                     {
-                        int toSend = Math.Min(sendWindow, memory.Length);
-                        toSend = Math.Min(toSend, RemoteMaxPacketSize);
-                        if (Interlocked.CompareExchange(ref _sendWindow, sendWindow - toSend, sendWindow) == sendWindow)
+                        int toSend = await _sendWindow.AquireAsync(aquireCount: memory.Length, exactCount: false, ChannelStopped, ct).ConfigureAwait(false);
+                        await this.SendChannelDataMessageAsync(memory.Slice(0, toSend), ct).ConfigureAwait(false);
+                        memory = memory.Slice(toSend);
+                        if (memory.IsEmpty)
```
HoundThe, line 361: Isn't the condition unnecessary?

src/Tmds.Ssh/MultiSemaphore.cs
```diff
@@ -0,0 +1,88 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Tmds.Ssh
+{
+    sealed class MultiSemaphore : IDisposable
```
HoundThe, line 10: Isn't there a Producer/Consumer collection implementation like https://docs.microsoft.com/cs-cz/dotnet/standard/collections/thread-safe/blockingcollection-overview that might do the same thing?

src/Tmds.Ssh/SshClient.ChannelContext.cs
tmds: :+1: I've removed it.

src/Tmds.Ssh/MultiSemaphore.cs
tmds: Those need real elements. We track space available for send. We don't have something to store.

PR #69, opened 2020-04-29
Sftp client base draft 

HoundThe:


src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
```diff
@@ -182,5 +182,55 @@ byte      SSH_MSG_CHANNEL_WINDOW_ADJUST
                 return packet.Move();
             }
         }
+
+        public static ValueTask StartSftpAsync(this ChannelContext context, CancellationToken ct)
```
tmds, line 186: Rename to `SendChannelSubsystemMessageAsync` and accept the subsystem as a string.

src/Tmds.Ssh/ChannelContextReceiveMessageExtensions.cs
```diff
@@ -67,5 +68,40 @@ static void ParseChannelOpenConfirmation(ReadOnlyPacket packet, string failureMe
                 }
             }
         }
+
+        public static async ValueTask<int> ReceiveServerVersionAsync(this ChannelContext context, string failureMessage, CancellationToken ct)
```
tmds, line 72: Let's move this into `SftpClient`.

src/Tmds.Ssh/ChannelContextReceiveMessageExtensions.cs
```diff
@@ -67,5 +68,40 @@ static void ParseChannelOpenConfirmation(ReadOnlyPacket packet, string failureMe
...
+            */
+
+            static int ParseSftpVersion(ReadOnlyPacket packet, string failureMessage)
+            {
+                var reader = packet.GetReader();
+                reader.ReadMessageId(MessageId.SSH_MSG_CHANNEL_DATA);
+                // Also maybe deal with multiple packets sent with the VERSION packet?
+                // var channelId = reader.ReadUInt32();
+                // var dataLength = reader.ReadUInt32();
+                // var sftpPacketLength = reader.ReadUInt32();
```
tmds, line 96: Get rid of the comments.

src/Tmds.Ssh/ChannelContextReceiveMessageExtensions.cs
```diff
@@ -67,5 +68,40 @@ static void ParseChannelOpenConfirmation(ReadOnlyPacket packet, string failureMe
...
+                reader.Skip(12);
+                var type = (SftpPacketType)reader.ReadByte();
+                if (type != SftpPacketType.SSH_FXP_VERSION)
+                    ThrowHelper.ThrowProtocolUnexpectedSftpPacketType(SftpPacketType.SSH_FXP_VERSION);
```
tmds, line 100: Let's add a method to the reader, similar to the implementation of `ReadMessageId`.

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
```diff
@@ -182,5 +182,55 @@ byte      SSH_MSG_CHANNEL_WINDOW_ADJUST
...
+                writer.WriteMessageId(MessageId.SSH_MSG_CHANNEL_REQUEST);
+                writer.WriteUInt32(context.RemoteChannel);
+                writer.WriteString("subsystem");
+                writer.WriteBoolean(true);
+                writer.WriteString(subsystem);
+                return packet.Move();
+            }
+        }
+
+        public static ValueTask SftpInitMessageAsync(this ChannelContext context, uint version, CancellationToken ct)
```
tmds, line 211: Let's move this also into the `SftpClient`.

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
```diff
@@ -182,5 +182,55 @@ byte      SSH_MSG_CHANNEL_WINDOW_ADJUST
...
+        {
+            return context.SendPacketAsync(CreatePacket(context, version), ct); // TODO later build sftp over SendChannelDataMessageAsync()
```
tmds, line 213: And indeed, we should use `SendChannelDataMessageAsync`.

src/Tmds.Ssh/LoggingExtensions.cs
```diff
@@ -180,7 +180,7 @@ public PacketPayload(ReadOnlyPacket packet)
 
             public override string ToString()
             {
-                const int maxDataLength = 2 * PrettyBytePrinter.BytesPerLine;
+                const int maxDataLength = 20 * PrettyBytePrinter.BytesPerLine;
```
tmds, line 183: Should we update this to `20`?

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Buffers;
+using System.Buffers.Binary;
+
+namespace Tmds.Ssh
+{
+    public class SftpFile
+    {
+        private byte[] handle;
```
tmds, line 11: handle -> _handle.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Buffers;
+using System.Buffers.Binary;
+
+namespace Tmds.Ssh
+{
+    public class SftpFile
```
tmds, line 9: Let's make this `sealed`.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        private byte[] handle;
+        internal SftpFile(byte[] handle)
+        {
+            this.handle = handle;
+        }
+    }
```
tmds, line 16: To make it a bit more interesting, let's also implement the `ValueTask CloseAsync();` method on this type.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
...
+        Write = 0x00000002,
+        Append = 0x00000004,
+        CreateNewOrOpen = 0x00000008,
+        Truncate = 0x00000010,
+        CreateNew = 0x00000020 | CreateNewOrOpen,
+    }
+
+    public partial class SftpClient
+    {
+        public async Task<SftpFile> OpenFileAsync(string path, SftpOpenFlags openFlags)
```
tmds, line 30: Let's change the return type to `ValueTask<SftpFile>`.
I think you can just update the return type, and it may work.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
...
+            writer.WriteByte((byte)SftpPacketType.SSH_FXP_OPEN);
+            writer.WriteUInt32(0);
+            writer.WriteString(filename);
+            writer.WriteUInt32((int)flags);
+            writer.WriteUInt32(0);
+            return packet.Move();
+        }
+    }
+
+    class OpenFileOperation : SftpOperation
```
tmds, line 70: Let's make this one `sealed`.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,108 @@
...
+            await SendRequestAsync(packet.Move(), operation);
+
+            return await operation.Task;
+
+            // int requestId = GetNextRequestId();
+
+            // _operations.TryAdd(requestId, operation);
+
+            // await _context.SftpOpenFileMessageAsync((UInt32)requestId, path, openFlags, attributes, default);
+            // return await operation.Task;
```
tmds, line 44: We can clean up these comments.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,156 @@
...
+
+namespace Tmds.Ssh
+{
+    abstract class SftpOperation
+    {
+        public abstract ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<byte> fields);
+
+        protected static Exception CreateExceptionForStatus(ReadOnlySequence<byte> fields)
+        {
+            return new SshException("TODO"); // TODO: use fields
```
tmds, line 16: The TODO is about adding a `ParseStatusFields` method that parses  https://tools.ietf.org/html/draft-ietf-secsh-filexfer-13#section-9.1.
Let's read
```
       uint32 error/status code
       string error message (ISO-10646 UTF-8 [RFC-2279])
```
We can add an enum for the status code: e.g. `SftpErrorCode` with the error codes from that section.
The `ErrorCode` can be added as a property on the `SshException`.
We can use error message as the `Message` for the `SshException`.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,156 @@
...
+        private ConcurrentDictionary<uint, SftpOperation> _operations;
+
+        private ValueTask SendRequestAsync(Packet packet, SftpOperation operation)
+        {
+            /*
+                0: SSH_MSG_CHANNEL_DATA  (1)
+                1: RemoteChannel         (4)
+                5: DATA length           (4)
+                9: SFTP length           (4)
+                13: SFTP packet type     (1) ] // Already filled in.
```
tmds, line 45: Can you remove the `]`?

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,156 @@
...
+        // SSH_MSG_CHANNEL_DATA          (1)
+        // RemoteChannel                 (4)
+        // SSH_MSG_CHANNEL_DATA length   (4)
+        // SFTP length                   (4)
+        const int DataHeaderLength = 13;
+        const int SftpVersion = 3;
+        private readonly ChannelContext _context;
+        private readonly Task _receiveLoopTask;
+        private int _requestId;
+        private ConcurrentDictionary<uint, SftpOperation> _operations;
```
tmds, line 36: This can be `readonly`.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,156 @@
...
+            return new SshException("TODO"); // TODO: use fields
+        }
+
+        protected static Exception CreateExceptionForUnexpectedType(SftpPacketType type)
+        {
+            return new ProtocolException($"Unexpected response type: {type}");
+        }
+    }
+
+    public partial class SftpClient : IDisposable
```
tmds, line 25: We should make this `sealed`.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,156 @@
...
+                    {
+                        // Nothing yet
+                    }
+
+                } while (messageId != MessageId.SSH_MSG_CHANNEL_CLOSE);
+
+            }
+            catch (Exception e)
+            {
+                throw e;
```
tmds, line 109: Let's add a TODO for dealing this exception.
The handling should avoid new operations being started, and an exception to be thrown for all on-going requests.

src/Tmds.Ssh/SshClient.cs
```diff
@@ -45,7 +46,7 @@ public SshClient(string destination, Action<SshClientSettings>? configure = null
         {
             _abortCts = new CancellationTokenSource();
 
-            _settings = CreateSettingsForDetination(destination);
+            _settings = CreateSettingsForDetination(destination); // Isn't this a typo?
```
tmds, line 49: Yep typo, you can correct it.

src/Tmds.Ssh/SshClient.cs
```diff
@@ -676,6 +677,36 @@ private void FreeChannel(uint nr)
             _allocatedChannels[i] = _allocatedChannels[i] & ~mask;
         }
 
+        public async Task<SftpClient> OpenSftpClientAsync(CancellationToken ct)
```
tmds, line 680: Can you move this into it's own file (like SshClient.ExecCommand)?

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -32,7 +32,7 @@ internal SshClientSettings(string userName, string host, int port)
         internal string UserName { get; set; }
         internal string Host { get; set; }
         internal int Port { get; set; } = 22;
-        internal ILogger? Logger { get; set; } = null; // TODO: decide how to expose the logger publically.
+        public ILogger? Logger { get; set; } = null; // TODO: decide how to expose the logger publically.
```
tmds, line 35: Keep this one internal.

src/Tmds.Ssh/ThrowHelper.cs
```diff
@@ -33,6 +33,12 @@ public static void ThrowProtocolUnexpectedMessageId(MessageId messageId)
             throw new ProtocolException($"Unexpected message id: {messageId}.");
         }
 
+        [DoesNotReturn]
+        public static void ThrowProtocolUnexpectedSftpPacketType(SftpPacketType packetType)
+        {
+            throw new ProtocolException($"Unexpected SFTP packet type: {packetType}.");
+        }
+
```
tmds, line 41: Can you add a separate method in `ThrowProtocolUnexpectedSftpPacketType` that does the `new ProtocolException($"Unexpected SFTP packet type: {packetType}.");`
And call that for `CreateExceptionForUnexpectedType`, and `ThrowHelper.ThrowProtocolUnexpectedSftpPacketType`.

src/Tmds.Ssh/SshClient.cs
HoundThe: How should I name it? SshClient.SftpSubsystem ?

src/Tmds.Ssh/LoggingExtensions.cs
HoundThe: No, I guess not. It was just for my personal use, removed

src/Tmds.Ssh/ThrowHelper.cs
HoundThe: You mean making a function that does the throwing inside the Throw.. function and calling that?

src/Tmds.Ssh/SequenceReader.cs
```diff
@@ -78,6 +78,19 @@ public void ReadMessageId(MessageId expectedValue)
                 ThrowHelper.ThrowProtocolUnexpectedValue();
             }
         }
+        public SftpPacketType ReadSftpPacketType()
```
tmds, line 81: nit: newline above this

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using System.Buffers;
+using System.Buffers.Binary;
+
+namespace Tmds.Ssh
+{
+    public sealed class SftpFile
+    {
+        private byte[] _handle;
```
tmds, line 11: `readonly` field, and add a newline below.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,214 @@
...
+                        string          extension-data
+            */
+
+            static int ParseSftpVersion(ReadOnlyPacket packet, string failureMessage)
+            {
+                var reader = packet.GetReader();
+                reader.ReadMessageId(MessageId.SSH_MSG_CHANNEL_DATA);
+                reader.Skip(12);
+                reader.ReadSftpPacketType(SftpPacketType.SSH_FXP_VERSION);
+                var version = (int)reader.ReadUInt32();
```
tmds, line 183: let's return version as a `uint` from this method.

src/Tmds.Ssh/SshClient.cs
tmds: Let's name it `SshClient.Sftp`.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,214 @@
...
+            return _context.SendChannelDataAsync(packet, default(CancellationToken)); // TODO: cancellation
+        }
+
+        internal SftpClient(ChannelContext context)
+        {
+            _context = context;
+            _operations = new ConcurrentDictionary<uint, SftpOperation>();
+        }
+
+        public async Task InitAsync(CancellationToken ct)
```
tmds, line 68: `InitAsync` should be `internal`.

src/Tmds.Ssh/SshClient.cs
```diff
@@ -676,6 +677,36 @@ private void FreeChannel(uint nr)
...
+                //     await context.SftpInitMessageAsync(3, ct).ConfigureAwait(false);
+                //     // TODO add server negotiation in case server would have min. version < 3, but was able to do a version 3 aswell
+                //     int serverVersion = await context.ReceiveServerVersionAsync("Failed to negotiate SFTP", ct).ConfigureAwait(false);
+                // }
+
+                return new SftpClient(context);
+            }
+            catch
+            {
+                context.Dispose(); // closing channel
```
tmds, line 706: We need to dispose the sftp client in case we've constructed it, similar to:
https://github.com/tmds/Tmds.Ssh/blob/10e93706c323b4ba543da8adb681bc550ad6bd41/src/Tmds.Ssh/SshClient.ExecCommand.cs#L56-L67

src/Tmds.Ssh/ThrowHelper.cs
tmds:
```cs
        [DoesNotReturn]
        public static void ThrowProtocolUnexpectedSftpPacketType(SftpPacketType packetType)
        {
            throw new CreateNewProtocolUnexpectedSftpPacketType(packetType);
        }

        public static Exception CreateNewProtocolUnexpectedSftpPacketType(SftpPacketType packetType)
        {
            return new ProtocolException($"Unexpected SFTP packet type: {packetType}.");
        }
```
and call `CreateNewProtocolUnexpectedSftpPacketType` from `CreateExceptionForUnexpectedType`.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,214 @@
...
+        {
+            await SftpInitMessageAsync(3, ct).ConfigureAwait(false);
+            // TODO add server negotiation in case server would have min. version < 3, but was able to do a version 3 aswell
```
tmds, line 71: Is there a TODO here? It's ok if we only work with servers that support version 3 and above.
Since we've sent 3, those servers should reply with 3 also:

>    'version' is the lower of the protocol version supported by the
   server and the version number received from the client.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,214 @@
...
+                    {
+                        // Nothing yet
+                    }
+
+                } while (messageId != MessageId.SSH_MSG_CHANNEL_CLOSE);
+
+            }
+            catch (Exception e)
+            {
+                // TODO Handling
```
tmds, line 109: This TODO may be left open.

src/Tmds.Ssh/ThrowHelper.cs
tmds: Actually, we can remove `ThrowProtocolUnexpectedSftpPacketType`. It's no longer being used.

src/Tmds.Ssh/SftpClient.cs
HoundThe: Yes I've misunderstood the negotiation, for some reason I thought they both will supply its on lowest spec support when I've read it first time. But that doesn't make sense, will remove

src/Tmds.Ssh/SshClient.cs
HoundThe: But only thing the SftpClient would dispose is the context right? Which we already do

HoundThe: I guess with the added init there, its true we might do more than channel disposing, maybe of the ReceiveLoopAsync?

src/Tmds.Ssh/SftpClient.File.cs
HoundThe: Hmm, should the method actually send requests to close the handle? How should that be implemented outside of the SftpClient?

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
HoundThe: For all SFTP or just these starting messages?

HoundThe: using SendRequestAsync in the SftpClient directly isn't enough? :thinking:

src/Tmds.Ssh/SftpClient.File.cs
HoundThe: I've implemented something for now

src/Tmds.Ssh/SftpClient.cs
HoundThe: Okey, so for a status success do nothing and any error results in an exception

HoundThe: Also you are linking a draft to version 6, but I think we're doing version 3 now?

HoundThe: But what if we don't want to throw exception on every STATUS message? maybe parse fields first and if its not success then pass it to the CreateExceptionForStatus?

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
```diff
@@ -28,8 +28,9 @@ byte      SSH_MSG_CHANNEL_FAILURE
             }
         }
 
-        public static ValueTask SendChannelDataMessageAsync(this ChannelContext context,  ReadOnlyMemory<byte> memory, CancellationToken ct)
+        public static ValueTask SendChannelDataMessageAsync(this ChannelContext context, ReadOnlyMemory<byte> memory, CancellationToken ct)
         {
+            // TODO Use SendChannelDataAsync because of a Window?
```
tmds, line 33: Good catch. This is a bug!

src/Tmds.Ssh/SftpClient.File.cs
tmds: We don't want `SftpClient` as an argument here, can you make it a field that gets set in the constructor?

src/Tmds.Ssh/SftpClient.cs
tmds: Yes, that makes sense. Now we have such a case with SftpFile.CloseAsync.

test/Tmds.Ssh.TestApp/Program.cs
```diff
@@ -52,6 +52,11 @@ static async Task<int> Main(string[] args)
 
                 Console.WriteLine("Process exited with exit code " + remoteProcess.ExitCode);
             }
+            else if (command == "sftp")
+            {
+                using var sftpClient = await client.OpenSftpClientAsync(default);
+                var file = await sftpClient.OpenFileAsync("/tmp/test", SftpOpenFlags.CreateNewOrOpen);
```
tmds, line 58: Can you add a call to CloseAsync?

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
tmds: The init packet is a bit special, because it doesn't have a request id.

src/Tmds.Ssh/SshClient.cs
tmds: I think we'll only do Dispose of ChannelContext.
It's a style preference, once we've created the SftpClient, I feel like it own's the ChannelContext so we should Dispose that instead of disposing ChannelContext directly.

tmds: (After SftpClient was used by the user, Dispose will have more side-effects, like waiting for the ReceiveLoop)

HoundThe: https://github.com/tmds/Tmds.Ssh/blob/fa03344365d303ce44dd7976e179343549ad931f/src/Tmds.Ssh/SshClient.Sftp.cs#L21-L33

But because we moved InitAsync there, the dispose should already be able to dispose the ReceiveLoop

src/Tmds.Ssh/SshClient.Sftp.cs
```diff
@@ -0,0 +1,37 @@
...
+
+                sftpClient = new SftpClient(context);
+                await sftpClient.InitAsync(ct);
+
+                return sftpClient;
+            }
+            catch
+            {
+                if (sftpClient != null)
+                    sftpClient.Dispose();
```
tmds, line 31: write this as: `sftpClient?.Dispose();`

Can you refactor the code in SshClient.ExecCommand.cs to use the same control flow?

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,160 @@
...
+            var operation = new OpenFileOperation();
+
+            await SendRequestAsync(packet.Move(), operation);
+
+            return await operation.Task;
+        }
+
+        // TODO add CancellationToken
+        // This can be used for any handle, so also for directories
+        public async ValueTask<bool> CloseFileAsync(byte[] handle)
```
tmds, line 47: Make this `internal`, and rename it to `CloseHandleAsync`.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,160 @@
...
+        {
+            using var packet = CreateCloseMessage(handle);
```
tmds, line 49: Can you make the `CreateCloseMessage` a local function?

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,160 @@
...
+        Truncate = 0x00000010,
+        CreateNew = 0x00000020 | CreateNewOrOpen,
+    }
+
+    public partial class SftpClient
+    {
+        // TODO add CancellationToken
+        public async ValueTask<SftpFile> OpenFileAsync(string path, SftpOpenFlags openFlags)
+        {
+            using var packet = CreateOpenMessage(path, openFlags);
```
tmds, line 37: Make `CreateOpenMessage` a local function here too.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,160 @@
...
+    {
+        private readonly byte[] _handle;
+
+        internal SftpFile(byte[] handle)
+        {
+            _handle = handle;
+        }
+
+        // For now I am taking a client as an argument because not sure how to do this
+        public ValueTask<bool> CloseAsync(SftpClient client) => client.CloseFileAsync(_handle);
```
tmds, line 19: To get rid of the argument here, you need to pass `SftpClient` to `OpenFileOperation` as an argument, which can then pass it to `SftpFile` constructor.

src/Tmds.Ssh/SshClient.Sftp.cs
HoundThe: I guess you mean refactor this to use the same control flow as ExecCommand?

HoundThe: If so, then it's done

src/Tmds.Ssh/SftpClient.File.cs
HoundThe: Yes, I forgot to push the changes, done

src/Tmds.Ssh/SshClient.Sftp.cs
tmds: Either is fine, as long as it is consistent.

src/Tmds.Ssh/SshClient.Sftp.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Tmds.Ssh
+{
+    public sealed partial class SshClient : IDisposable
+    {
+
+        public async Task<SftpClient> OpenSftpClientAsync(CancellationToken ct)
```
tmds, line 10: nit: remove newline

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,205 @@
...
+
+        private ValueTask SftpInitMessageAsync(uint version, CancellationToken ct)
+        {
+            /*
+                    uint32 	length
+                    byte 	type
+                    uint32 	version
+            */
+
+            var memory = new Memory<byte>(new byte[9]);
```
tmds, line 195: Can you implement this similar to the methods in `ChannelContextSendMessageExtensions.cs` (using `context.RentPacket();` etc)?

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,161 @@
...
+    sealed class CloseHandleOperation : SftpOperation
+    {
+        private TaskCompletionSource<bool> _tcs = new TaskCompletionSource<bool>();
+
+        // TODO can return status and on some systems even close can fail
+        public override ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<byte> fields, SftpClient client)
+        {
+            if (type != SftpPacketType.SSH_FXP_STATUS)
+            {
+                _tcs.SetException(CreateExceptionForStatus(fields));
```
tmds, line 148: should be `CreateExceptionForUnexpectedType`

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+            else
+            {
+                // if (SSH_FXP_STATUS == success) else.....
+                _tcs.SetResult(true); // TODO parsing of SSH_FXP_STATUS
```
tmds, line 153: This can use the same method as `CreateExceptionForStatus` to parse status fields.
The method should be in `SftpOperation` and have a signature that looks like:

```
protected static (SftpErrorCode errorCode, string? errorMessage) ParseStatusFields(ReadOnlySequence<byte> fields)
```

src/Tmds.Ssh/SshClient.ChannelContext.cs
```diff
@@ -349,6 +349,27 @@ internal void DoDispose()
                 }
             }
 
+            public override async ValueTask SendChannelDataAsync(Packet packet, CancellationToken ct)
+            {
+                using var pkt = packet.Move();
+                long toSend = packet.PayloadLength - 9; // TODO: verify
```
tmds, line 355: You can remove this TODO. `9` is the correct value. It is the offset of the data in the `SSH_MSG_CHANNEL_DATA` payload.

src/Tmds.Ssh/SftpClient.cs
HoundThe: That's what was there before, but in the comment above you said to use the SendChannelDataMessageAsync, that takes Memory<T>, and I am not sure how to create Memory<T> from the packet to use the method :thinking:

tmds: I meant ChannelContext.SendChannelData~~Message~~Async which takes a Packet arg.

HoundThe: Oh ye, that makes sense

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -145,12 +145,20 @@ public override ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<b
         {
             if (type != SftpPacketType.SSH_FXP_STATUS)
```
tmds, line 146: nit: can you swap the blocks by inverting the condition.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -145,12 +145,20 @@ public override ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<b
...
             {
-                _tcs.SetException(CreateExceptionForStatus(fields));
+                _tcs.SetException(CreateExceptionForUnexpectedType(type));
             }
             else
             {
-                // if (SSH_FXP_STATUS == success) else.....
-                _tcs.SetResult(true); // TODO parsing of SSH_FXP_STATUS
+                var statusTuple = ParseStatusFields(fields);
```
tmds, line 152: I'm using the same coding conventions as .NET Core, which means only `var` when the type is explicit on the rhs (like constructor).
`(SftpErrorCode errorCode, string? errorMessage)`

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
...
         public abstract ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<byte> fields, SftpClient client);
 
+        protected static Exception CreateExceptionForStatus(SftpErrorCode errorCode, string? errorMessage)
+        {
+            return new SshException(errorMessage ?? "No error message specified", errorCode);
+        }
         protected static Exception CreateExceptionForStatus(ReadOnlySequence<byte> fields)
         {
-            return new SshException("TODO"); // TODO: use fields
+            var statusTuple = ParseStatusFields(fields);
```
tmds, line 20: `(SftpErrorCode errorCode, string? errorMessage)`

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
...
+            return new SshException(statusTuple.errorMessage ?? "No error message specified", statusTuple.errorCode);
```
tmds, line 21: `"No error message specified"` -> `$"Error {errorCode}"`

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
...
         }
 
         protected static Exception CreateExceptionForUnexpectedType(SftpPacketType type)
         {
             return new ProtocolException($"Unexpected response type: {type}");
         }
+        protected static (SftpErrorCode errorCode, string? errorMessage) ParseStatusFields(ReadOnlySequence<byte> fields)
```
tmds, line 28: nit: add a space above this

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
     {
         public abstract ValueTask HandleResponse(SftpPacketType type, ReadOnlySequence<byte> fields, SftpClient client);
 
+        protected static Exception CreateExceptionForStatus(SftpErrorCode errorCode, string? errorMessage)
+        {
+            return new SshException(errorMessage ?? "No error message specified", errorCode);
+        }
```
tmds, line 17: nit: add a newline below

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
...
 
         protected static Exception CreateExceptionForUnexpectedType(SftpPacketType type)
         {
             return new ProtocolException($"Unexpected response type: {type}");
         }
+        protected static (SftpErrorCode errorCode, string? errorMessage) ParseStatusFields(ReadOnlySequence<byte> fields)
+        {
+            var reader = new SequenceReader(fields);
+            var errorCode = (SftpErrorCode)reader.ReadUInt32();
+            var errorMessage = reader.ReadUtf8String();
```
tmds, line 32:
```
string? errorMessage = reader.ReadUtf8String();
if (errorMessage.Length == 0)
{
    errorMessage = null;
}
```

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -11,15 +11,28 @@ abstract class SftpOperation
...
+            return new SshException(statusTuple.errorMessage ?? "No error message specified", statusTuple.errorCode);
         }
 
         protected static Exception CreateExceptionForUnexpectedType(SftpPacketType type)
         {
             return new ProtocolException($"Unexpected response type: {type}");
         }
+        protected static (SftpErrorCode errorCode, string? errorMessage) ParseStatusFields(ReadOnlySequence<byte> fields)
+        {
+            var reader = new SequenceReader(fields);
```
tmds, line 30: Can you add a comment above this:
```
/*
        uint32     error/status code
        string     error message (ISO-10646 UTF-8 [RFC-2279])
        string     language tag (as defined in [RFC-1766])
*/
```

src/Tmds.Ssh/SftpClient.cs
tmds: We should have a dedicated exception type `SftpException : SshException`.
The constructor should accept `(SftpErrorCode errorCode, string? errorMessage)`.

src/Tmds.Ssh/SshException.cs
```diff
@@ -9,7 +9,12 @@ namespace Tmds.Ssh
     [Serializable]
     public class SshException : Exception
     {
+        public SftpErrorCode ErrorCode { get; internal set; }
```
tmds, line 12: We'll move this to the `SftpException` type.

src/Tmds.Ssh/SftpClient.File.cs
```diff
@@ -0,0 +1,169 @@
...
+            return handle;
+        }
+    }
+
+    // This can be used for any handle, so also for directories
+    sealed class CloseHandleOperation : SftpOperation
+    {
+        private TaskCompletionSource<bool> _tcs = new TaskCompletionSource<bool>();
+
+        // TODO can return status and on some systems even close can fail
```
tmds, line 143: Can this TODO be removed, or is there still something to be tackled?

HoundThe: Removed

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
tmds: Can you move this method in the `ChannelContext` class and make it `private`?

src/Tmds.Ssh/SftpException.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Tmds.Ssh
+{
+    [Serializable]
+    public class SftpException : Exception
+    {
+        public SftpErrorCode ErrorCode { get; internal set; }
+		
+        internal SftpException(string message) : base(message) { }
+        internal SftpException(string message, SftpErrorCode errorCode) : base(message)
```
tmds, line 14: Can you make this `(SftpErrorCode errorCode, string? message)` and handle `null` message here?
You can remove the `SftpException(string message)` constructor.

src/Tmds.Ssh/SftpException.cs
```diff
@@ -0,0 +1,20 @@
...
+// See file LICENSE for full license details.
+
+using System;
+
+namespace Tmds.Ssh
+{
+    [Serializable]
+    public class SftpException : Exception
+    {
+        public SftpErrorCode ErrorCode { get; internal set; }
```
tmds, line 11: We don't need the `internal set`.

src/Tmds.Ssh/SftpException.cs
```diff
@@ -0,0 +1,20 @@
...
+		
+        internal SftpException(string message) : base(message) { }
+        internal SftpException(string message, SftpErrorCode errorCode) : base(message)
+        {
+            ErrorCode = errorCode;
+        }
```
tmds, line 17: nit: add a newline below.

src/Tmds.Ssh/SftpException.cs
```diff
@@ -0,0 +1,20 @@
...
+        internal SftpException(string message, Exception inner) : base(message, inner) { }
```
tmds, line 18: You can remove this constructor too.

src/Tmds.Ssh/SftpClient.cs
```diff
@@ -0,0 +1,234 @@
...
+                    using var packet = await _context.ReceivePacketAsync(ct: default).ConfigureAwait(false);
+                    messageId = packet.MessageId!.Value;
+
+                    if (messageId == MessageId.SSH_MSG_CHANNEL_DATA)
+                    {
+                        await HandleChannelData(packet.Move()).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        // Nothing yet
```
tmds, line 121: Can you remove this empty block.

src/Tmds.Ssh/ChannelContextSendMessageExtensions.cs
HoundThe: Not sure if I understand you, ChannelContext is abstract class, so I guess you mean moving into  SshChannelContext, but I had to get rid of the static qualifiers

tmds:
This looks ready to merge. Nice work!
Can you add the licence header to all new files:
```
// This file is part of Tmds.Ssh which is released under MIT.
// See file LICENSE for full license details.

```

HoundThe:
Nice! Header is added

PR #88, opened 2021-04-06
Build libssh for win-x64 in container

tmds:


src/libssh.win-x64/context-dir/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+    tar -xf libssh-*.tar.xz && \
+    ls && \
+    mv $(find . -maxdepth 1 -name 'libssh-*' -type d) libssh
+
+# Build and install libssh.
+RUN mkdir libssh/build && \
+    cd libssh/build && \
+    mingw64-cmake .. && \
+    make -j$(nproc) && \
+    make install
```
tmds, line 21: @omajid I'm getting the Fedora libssh sources here and then use `mingw64-cmake` to build them for `win-x64`. Do you have suggestions on how I can do this better?

I'm just building stuff at this point, I haven't verified these dlls actually work on Windows.

You can also look at this later and send a PR to improve how things get built.

Issue #182, opened 2024-06-10
Support custom subsystem with ExecuteAsync


jborean93:
It would be great if `ExecuteAsync` (or some variant) had a way to specify using a custom subsystem rather than a command. For example the OpenSSH cli can open a connection to a subsystem with `ssh -s user@remote subsystem_name`.

As some background PowerShell Remoting is done through a custom subsystem configuration https://learn.microsoft.com/en-us/powershell/scripting/security/remoting/ssh-remoting-in-powershell?view=powershell-7.4. The SSH service registers a subsystem (typically called `powershell`) that is set to run a set executable with an argument.

```
Subsystem powershell c:/progra~1/powershell/7/pwsh.exe -sshs -nologo
```

The PowerShell client internally runs uses `ssh -s user@remote powershell` to open the connection and exchange the PSRemoting packets over stdin/stdout of this new connection. I'm hoping to use this library to achieve the same thing instead of relying on the `ssh` binary.

tmds:
Thanks for creating the issue, and elaborating on the use-case!

What do you think about this API:

```cs
Task<RemoteProcess> ExecuteSubystemAsync(string subsystem CancellationToken cancellationToken);
Task<RemoteProcess> ExecuteSubystemAsync(string subsystem, ExecuteOptions? options = null, CancellationToken cancellationToken = default);
```

jborean93:
You read my mind as I was working on https://github.com/tmds/Tmds.Ssh/pull/183 which used that exact name!

PR #183, opened 2024-06-10
Add ExecuteSubsystemAsync for subsystem based proc

jborean93:
Adds the ExecuteSubsystemAsync call which can create an SSH channel from a subsystem and return the RemoteProcess interface. This is used when the SSH server defines custom subsystems in its sshd_config file that allows a client to create a channel using a pre-defined subsystem name rather than through an explicit command line argument.

Fixes: https://github.com/tmds/Tmds.Ssh/issues/182

### DIFF
--- a/src/Tmds.Ssh/ISshClientImplementation.cs
+++ b/src/Tmds.Ssh/ISshClientImplementation.cs
@@ -11,6 +11,7 @@ interface ISshClientImplementation
 {
     Task ConnectAsync(CancellationToken cancellationToken);
     Task<ISshChannel> OpenRemoteProcessChannelAsync(Type channelType, string command, CancellationToken cancellationToken);
+    Task<ISshChannel> OpenRemoteSubsystemChannelAsync(Type channelType, string subsystem, CancellationToken cancellationToken);
     Task<ISshChannel> OpenTcpConnectionChannelAsync(Type channelType, string host, int port, CancellationToken cancellationToken);
     Task<ISshChannel> OpenUnixConnectionChannelAsync(Type channelType, string path, CancellationToken cancellationToken);
     Task<ISshChannel> OpenSftpClientChannelAsync(Action<SshChannel> onAbort, CancellationToken cancellationToken);
--- a/src/Tmds.Ssh/SshSession.cs
+++ b/src/Tmds.Ssh/SshSession.cs
@@ -623,7 +623,7 @@ sealed partial class SshSession : ISshClientImplementation
     }
 
     public async Task<ISshChannel> OpenSftpClientChannelAsync(Action<SshChannel> onAbort, CancellationToken cancellationToken)
-        => await OpenSubsystemChannelAsync(typeof(SftpChannel), onAbort, "sft", cancellationToken).ConfigureAwait(false);
+        => await OpenSubsystemChannelAsync(typeof(SftpChannel), onAbort, "sftp", cancellationToken).ConfigureAwait(false);
 
     private async Task<ISshChannel> OpenSubsystemChannelAsync(Type channelType, Action<SshChannel>? onAbort, string subsystem, CancellationToken cancellationToken)
     {
--- a/src/Tmds.Ssh/SshSession.cs
+++ b/src/Tmds.Ssh/SshSession.cs
@@ -583,6 +583,9 @@ sealed partial class SshSession : ISshClientImplementation
         }
     }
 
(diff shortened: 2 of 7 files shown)


src/Tmds.Ssh/ISshClientImplementation.cs
```diff
@@ -11,6 +11,7 @@ interface ISshClientImplementation
 {
     Task ConnectAsync(CancellationToken cancellationToken);
     Task<ISshChannel> OpenRemoteProcessChannelAsync(Type channelType, string command, CancellationToken cancellationToken);
+    Task<ISshChannel> OpenRemoteSubsystemChannelAsync(Type channelType, string subsystem, CancellationToken cancellationToken);
```
jborean93, line 14: I'm unsure how you want to handle this, IIRC the default access modifier for an interface is `public` so this may be a breaking change API wise.

test/Tmds.Ssh.Tests/RemoteProcessTests.cs
```diff
@@ -44,6 +44,36 @@ public async Task HelloWorld(bool stderr)
...
+        using var process = await client.ExecuteSubsystemAsync(_sshServer.TestSubsystem);
+
+        await process.WriteLineAsync("echo -n 'hello world 1'");
+
+        byte[] buffer = new byte[512];
+        (bool isError, int bytesRead) = await process.ReadAsync(buffer, buffer);
+        Assert.False(isError);
+        Assert.Equal(helloWorld1Bytes, buffer.AsSpan(0, bytesRead).ToArray());
+
+        await process.WriteLineAsync("echo -n 'hello world 2'");
```
jborean93, line 62: I tried to do `>&2` to test out `isError` but it seems like SSH doesn't return `SSH_EXTENDED_DATA_STDERR` on subsystem based channels. I was unable to even get the `ssh` binary to return stderr on a pre-configured subsystem that wrote to stderr so I assume it's not something that can happen. I couldn't find anything in the RFC around this but I didn't look too hard.

src/Tmds.Ssh/ISshClientImplementation.cs
tmds: This interface itself is not public so there is no backwards compatibility concern.

In addition to the managed implementation, there once also was a [libssh](https://www.libssh.org) based backend. This interface is a reminiscent of that, and something to be removed.

fyi, C# has a feature to extend interfaces so it doesn't affect existing implementations, see [default interface methods](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/interface-implementation/default-interface-methods-versions).

test/Tmds.Ssh.Tests/RemoteProcessTests.cs
tmds: The test is good as is. It covers what we need.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -620,8 +623,11 @@ public async Task<ISshChannel> OpenUnixConnectionChannelAsync(Type channelType,
     }
 
     public async Task<ISshChannel> OpenSftpClientChannelAsync(Action<SshChannel> onAbort, CancellationToken cancellationToken)
+        => await OpenSubsystemChannelAsync(typeof(SftpChannel), onAbort, "sft", cancellationToken).ConfigureAwait(false);
```
tmds, line 626: typo: "sft" -> "sftp"

test/Tmds.Ssh.Tests/SshServer.cs
```diff
@@ -20,6 +20,7 @@ public class SshServer : IDisposable
     public string TestUserHome => $"/home/{TestUser}";
     public string TestUserPassword => "secret";
     public string TestUserIdentityFile => $"{ContainerBuildContext}/user_key_rsa";
+    public string TestSubsystem = "tmds_test";
```
tmds, line 23: nit: rename `tmds_test` to `test_subsystem`

jborean93: Has been renamed.

src/Tmds.Ssh/ISshClientImplementation.cs
jborean93: Thanks for clarifying

tmds:
Thanks for this contribution @jborean93!

Until this becomes available in a nuget.org release, you can use it from the CI feed: https://github.com/tmds/Tmds.Ssh?tab=readme-ov-file#ci-feed.

Issue #184, opened 2024-06-10
gssapi-with-mic Credential Support RFC 4462


jborean93:
I'm not aware what work is involved but it would be great if this supported the `gssapi-with-mic` auth protocol https://datatracker.ietf.org/doc/html/rfc4462. .NET has the [NegotiateAuthentication](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.negotiateauthentication?view=net-8.0) which wraps SSPI on Windows and GSSAPI on non-Windows which should hopefully offer the needed calls to do both authentication and wrapping that's required. I have a suspicion that it may not expose enough detail to achieve using the builtin class, like getting the MIC, or supporting delegation but maybe those are optional extras and .NET could expose a flag for those features in future versions.

jborean93:
Looks like .NET 9 added the MIC capabilities with https://github.com/dotnet/runtime/pull/96712 through [NegotiateAuthentication.ComputeIntegrityCheck](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.negotiateauthentication.computeintegritycheck?view=net-9.0#system-net-security-negotiateauthentication-computeintegritycheck(system-readonlyspan((system-byte))-system-buffers-ibufferwriter((system-byte)))). The only missing option that I would be interested in is delegation support but I can open an issue on the .NET runtime repo about that and it's an optional feature.

jborean93:
I've worked on this a bit and was able to get it working. Here is a diff with the basic changes needed. More work would need to be done to deal with error handling and more tidying up of the code. It also uses reflection for .NET 8 as the MIC method was not public until .NET 9.

<details>
  <summary>Expand to get diff</summary>

```diff
diff --git a/src/Tmds.Ssh/GssapiWithMicCredential.cs b/src/Tmds.Ssh/GssapiWithMicCredential.cs
new file mode 100644
index 0000000..b90e577
--- /dev/null
+++ b/src/Tmds.Ssh/GssapiWithMicCredential.cs
@@ -0,0 +1,253 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
+using System;
+using System.Buffers;
+using System.Linq;
+using System.Net;
+using System.Net.Security;
+using System.Reflection;
+using System.Security.Principal;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Tmds.Ssh;
+
[...]

PR #188, opened 2024-06-13
Add gssapi-with-mic user authentication

jborean93:
Add support for gssapi-with-mic user authentication. This can be used to connect to an SSH server that support GSSAPI/Kerberos authentication using GSSAPI on nix and SSPI on Windows.

Fixes: https://github.com/tmds/Tmds.Ssh/issues/184

Let me know what you think about my approach in the tests. Kerberos is very picky when it comes to hostnames but hacking in localhost seems to been enough for it to run locally. I'm also unsure how to deal with authentication errors so any suggestion as to when it should raise an exception or just ignore and try and next one would be nice to hear.


src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,282 @@
...
+    private MethodInfo GetMICMethod { get; }
+#endif
+
+    /// <summary>
+    /// Create a GSSAPI user authentication context with password string.
+    /// </summary>
+    /// <param name="password">The password to use for the GSSAPI username. Set to null to use a cached credential.</param>
+    /// <param name="delegateCredential">Request delegation on the GSSAPI context.</param>
+    /// <param name="serviceName">Override the service principal name (SPN), default it to use the connection hostname</param>
+    public GssapiWithMicCredential(string? password = null, bool delegateCredential = false, string? serviceName = null) : this(() => password, delegateCredential, serviceName)
```
tmds, line 44: suggested rename: `delegateCredential` to `forwardCredentials`.

tmds: suggested rename: `serviceName` to `serverIdentity`.

src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,282 @@
...
+    { }
+
+    /// <summary>
+    /// Create a GSSAPI user authentication context with a password prompt.
+    /// </summary>
+    /// <param name="passwordPrompt">A prompting function that is called to retrieve the password when needed.</param>
+    /// <param name="delegateCredential">Request delegation on the GSSAPI context.</param>
+    /// <param name="serviceName">Override the service principal name (SPN), default it to use the connection hostname</param>
+    public GssapiWithMicCredential(Func<string?> passwordPrompt, bool delegateCredential = false, string? serviceName = null)
```
tmds, line 53: > when needed

Looking at the implementation, it seems we're always prompting and not only when needed?

I think we should leave out the `passwordPrompt` and `password` arguments if that is the case.

test/Tmds.Ssh.Tests/SshServer.cs
```diff
@@ -247,6 +297,18 @@ public SshClientSettings CreateSshClientSettings(Action<SshClientSettings>? conf
     }
 }
 
+internal partial class Libc
```
tmds, line 300: Can you use https://github.com/tmds/Tmds.ExecFunction to run the test that set environment variables in a separate process?

tmds:
@jborean93 thanks for picking this up! The PR is already looking quite comprehensive!

I read through the PR and added some feedback.
I'll try to do an in-depth review in the coming week

src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,282 @@
...
+using System.Net.Security;
+using System.Reflection;
+using System.Security.Principal;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Tmds.Ssh;
+
+public sealed class GssapiWithMicCredential : Credential
```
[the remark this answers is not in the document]
jborean93: What I'm not sure on is how this would look if `gssapi-keyex` is ever implemented. The `keyex` method introduces a new host verification process bundled with user authentication. Technically GSSAPI can also negotiate more than just Kerberos but I'm not sure if any server implementation really does anything other than Kerberos through it.

src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,282 @@
...
+
+    // Kerberos - 1.2.840.113554.1.2.2
+    private static readonly byte[] KRB5_OID = [ 0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x12, 0x01, 0x02, 0x02 ];
+
+    private readonly Func<string?> _getPassword;
+    private readonly bool _delegateCredential;
+    private readonly string? _serviceName;
+
+#if NET8_0
+    private delegate void GetMICDelegate(ReadOnlySpan<byte> data, IBufferWriter<byte> writer);
```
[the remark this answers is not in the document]
jborean93: I'll definitely have a look!

src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,282 @@
...
+        _serviceName = serviceName;
+
+#if NET8_0
+        GetMICMethod = _getMicMethInfo.Value ?? throw new InvalidOperationException("Failed to find GetMIC method needed for .NET 8.0.");
+#endif
+    }
+
+    internal async Task<bool> TryAuthenticate(SshConnection connection, ILogger logger, SshClientSettings settings, SshConnectionInfo connectionInfo, CancellationToken ct)
+    {
+        string spn = _serviceName ?? $"host@{connectionInfo.Host}";
```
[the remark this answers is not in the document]
jborean93: It's the service principal name (SPN) that is used to identify what service the client wants to authenticate with on the KDC. This should match up with a registered SPN on the KDC that the OpenSSH implementation is running as. The `host` prefix is defined in the RFC https://www.rfc-editor.org/rfc/rfc4462#section-7.1 but theoretically an SSH service could run under any SPN. The RFC does have some notes around canonicalisation which the Kerberos library can be configured to do so if the user opts into it.

src/Tmds.Ssh/GssapiWithMicCredential.cs
jborean93: Up to you, I went with `delegateCredential` because the GSSAPI/SSPI flag has delegation in the name. The OpenSSH option for this is also `GSSAPIDelegateCredentials` to there is some consistency there (I can make it plural if you like).

For `serviceName` I settled on that because it refers to the Service Principal Name (SPN) which is a common term used in Kerberos. Happy to rename either though if you prefer that.

src/Tmds.Ssh/GssapiWithMicCredential.cs
jborean93: I mostly just followed the `PasswordCredential` setup (which also always prompts for a password). I assumed it was because a previous credential might be called before this one so the password may not have been prompted. Theoretically it can also fail if the server doesn't allow the auth OID that we are implementing (Kerberos).

It'll certainly be simpler to not have that and just accept a nullable string if that's what you would like here.

test/Tmds.Ssh.Tests/SshServer.cs
jborean93: That looks great, much better solution to using this, will try it out.

src/Tmds.Ssh/GssapiWithMicCredential.cs
tmds: > I think we should leave out the passwordPrompt and password arguments if that is the case.

Actually, we can keep the `password` arg.

Would it make sense to accept it as: `NetworkCredential? credential = null`?

src/Tmds.Ssh/GssapiWithMicCredential.cs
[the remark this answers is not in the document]
tmds: The `Credential` derived types are meant to capture information for authentication and (in theory) they should be usable in different situation. The "worst case" requires introducing a new type of credential.

I'd like a descriptive user-friendly name. I don't think `GssapiWithMic` means much to users of this library. I assume `KerberosCredential` does.

src/Tmds.Ssh/GssapiWithMicCredential.cs
tmds: > Up to you, I went with delegateCredential because the GSSAPI/SSPI flag has delegation in the name. The OpenSSH option for this is also GSSAPIDelegateCredentials to there is some consistency there (I can make it plural if you like).

Let's keep the OpenSSH based name.

> For serviceName I settled on that because it refers to the Service Principal Name (SPN) which is a common term used in Kerberos. Happy to rename either though if you prefer that.

If this a common Kerberos term. Let's also keep it.

src/Tmds.Ssh/GssapiWithMicCredential.cs
tmds: > I mostly just followed the PasswordCredential setup (which also always prompts for a password). I assumed it was because a previous credential might be called before this one so the password may not have been prompted. Theoretically it can also fail if the server doesn't allow the auth OID that we are implementing (Kerberos).

I would expect it to prompt only when needed. If it is always prompting, the user may provide the credential up-front.

For the password prompt, we actually skip the auth when the return value is `null`, while this is behaving different.

Let's leave out the `passwordPrompt` overload (for now).

src/Tmds.Ssh/GssapiWithMicCredential.cs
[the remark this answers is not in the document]
jborean93: Makes sense, Kerberos is certainly more user friendly.

src/Tmds.Ssh/GssapiWithMicCredential.cs
jborean93: > Let's leave out the passwordPrompt overload (for now).

Makes sense!

> Would it make sense to accept it as: NetworkCredential? credential = null?

While it certainly could be used it does have a weird setup where there's the connection username specified and the one in the `NetworkCredential`. Which one would be used if we accepted a `NetworkCredential`, what would happen if they differ?

tmds: I think we should use the one from the `NetworkCredential` and ignore the connection username.

tmds: I'm looking at the .NET runtime implementation here:

https://github.com/dotnet/runtime/blob/f52856334d4929839ef97b7c11376ac51ebc6b2e/src/libraries/System.Net.Security/src/System/Net/NegotiateAuthenticationPal.Unix.cs#L194-L196

It seems that it is only considering the `UserName` when the `Password` is also set.

So, I think this makes sense:

If no credential is specified, assume it to be `CredentialCache.DefaultNetworkCredentials` and don't send a username in the `gssapi-with-mic` message.

If a credential is specified, it must include a username and password.

wdyt?

tmds: > and don't send a username in the gssapi-with-mic message.

This depends on: _ The user name may be an empty string if it can be deduced from the
   results of the GSS-API authentication._. I'm not sure if this is the case?

tmds: > I'm not sure if this is the case?

I gave it a try (against the test you've included), and setting an empty username doesn't work.

We can do the following: if the NetworkCredential.UserName is an empty string, use the connection username, and otherwise use the NetworkCredential.UserName.

tmds: I just realized this username includes a domain too.

We can just start with an API that requires a `NetworkCredential`, and require a `UserName` to be set on this credential. The API can be relaxed later.

jborean93: > It seems that it is only considering the UserName when the Password is also set.

That is a good point. It is worth pointing out the path on Windows is a bit more convoluted but it calls [SspiEncodeStringsAsAuthIdentity](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-sspiencodestringsasauthidentity) which doesn't actually fail on an empty password but I think that's a pretty rare scenario to think about on Windows.

> I gave it a try (against the test you've included), and setting an empty username doesn't work.

While the RFC states that the username can be null I found that openssh-portable explicitly fails if the username is not set https://github.com/openssh/openssh-portable/blob/a8fbe2f7d0d96d299ee8e69769e3b51067978748/auth2-gss.c#L114-L118. This happens after the first message which includes the username and OID the client supports and before the authentication token is first generated.

The same also applies to the Win32-OpenSSH fork which has the same code. This effectively means that we cannot rely on the cached credential, you must supply a user in some shape or form.

> I just realized this username includes a domain too.

At least on Linux, the Kerberos libraries have the concept of a default realm if configured in the `krb5.conf`. This means that `Environment.UserName` could still be possible in some scenarios.

Ultimately I think what we could do is accept `NetworkCredential?` which if `null` will default to `CredentialCache.DefaultNetworkCredential` for the GSSAPI auth stage. The `username` string supplied in the SSH messages should use `credential.UserName` with a fallback to the `settings.UserName` so an explicit value is always provided. This has the added benefit of being able to run as a specific target user (which may be different from our kerb principal) but authenticate with either the default kerb ticket or the explicit kerb credentials provided. It also means a realm may not be needed on the SSH connection string username.

src/Tmds.Ssh/GssapiWithMicCredential.cs
```diff
@@ -0,0 +1,284 @@
...
+            string    user name (in ISO-10646 UTF-8 encoding)
+            string    service name (in US-ASCII)
+            string    "gssapi-with-mic" (US-ASCII method name)
+            uint32    n, the number of mechanism OIDs client supports
+            string[n] mechanism OIDs
+        */
+        using var packet = sequencePool.RentPacket();
+        var writer = packet.GetWriter();
+        writer.WriteMessageId(MessageId.SSH_MSG_USERAUTH_REQUEST);
+        writer.WriteString(userName);
```
tmds, line 176: Is it permitted to leave out the realm part when sending the username?

src/Tmds.Ssh/GssapiWithMicCredential.cs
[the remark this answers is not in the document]
tmds: Let's use `KerberosCredential`.

src/Tmds.Ssh/GssapiWithMicCredential.cs
jborean93: Yep, surprisingly the `username` portion in this step is the user sshd will map the authentication to. The SSH server verifies that this user is valid on the remote before even starting the Kerberos token exchange. This means that I can specify the target username as part of the SSH connection settings but use a completely different principal for authentication. On Linux as long as the Kerberos principal is mapped to that user using a `.k5login` file in the user's home dir it will log on as that user.

I added a test for this specific scenario https://github.com/tmds/Tmds.Ssh/pull/188/files#diff-ce2477da774fe35b103aac30b0d1ed4afd6e515ce3c5eb96ad791d09b4b2b27bR161. In this case both `testuser` and `testuser@REALM.TEST` is a valid user on the Linux target and both users have a `.k5login` file that map `testuser@REALM.TEST` to the local account allowing me to login as both users with the same Kerberos principal.

For Windows, the Windows host will automatically lookup the user in the realm it is joined to so this also works. I don't believe it has an analogue for mapping a Kerberos principal to any other account though so the username needs to match the Kerberos principal used.

Also as a side note, you can specify the connection username using a UPN string

```csharp
var settings = new SshClientSettings("server")
{
    UserName = "user@REALM.COM",
};
```

tmds: > For Windows, the Windows host will automatically lookup the user in the realm it is joined to so this also works. 

This is what I was expecting/hoping.

> The SSH server verifies that this user is valid on the remote before even starting the Kerberos token exchange.

Why does this check pass when we include a realm? What user is then checked against?

> Also as a side note, you can specify the connection username using a UPN string right now, it's just not possible to do so in the connection string due to the current parsing logic.

For this PR, I'd like to require an non-empty username when creating the `Credential`. And with https://github.com/tmds/Tmds.Ssh/issues/189 we can improve how it works with `SshClientSettings`.

jborean93: There are two different usernames that come into play with Kerberos authentication.

+ The SSH username that is supplied as part of the SSH messages
+ The Kerberos principal that is used with Kerberos auth

The first one is supplied as part of the initial OID message and the signed MIC data and is the username that the SSH server will attempt to map to a user on that host. For Linux this is typically just a user with the same name (what we do in the tests) but you can also use services like SSSD and maybe a PAM extension to support things like looking up the user through LDAP. For now all the server does is validates that this user is a valid user.

The second one is provided as part of the Kerberos authentication process. This is the Kerberos user that you are going to authenticate with and is aligned to either the cached credential retrieved by kinit on *nix, your logged in user on Windows, or the explicitly provided username/password. The server will validate that these credentials are valid for the principal specified.

Once the server has verified that the client is the Kerberos principal it will then determine if that principal can run as that user. On Windows I believe this just means that the Kerberos principal is the user specified in the SSH message, on *nix this is done by checking if the SSH username contains a `.k5login` file that contains the Kerberos principal used in the authentication.

In this PR these two usernames are derived with this logic:

+ The SSH user is either the Kerberos principal `UserName` if provided and not null/empty with a fallback to the connection setting `UserName`
+ The Kerberos principal is either used from the explicitly provided `NetworkCredential` or a cached credential

In the test case the container has two Linux users on the SSH host, `testuser` and `testuser@REALM.TEST`. Both accounts have a `.k5login` file that maps the Kerberos principal `testuser@REALM.TEST`. This allows us to login as the Linux user `testuser` or `testuser@REALM.TEST` using the same Kerberos principal `testuser@REALM.TEST`.

> For this PR, I'd like to require an non-empty username when creating the Credential. And with https://github.com/tmds/Tmds.Ssh/issues/189 we can improve how it works with SshClientSettings.

We can certainly do this but I don't think it's a good idea. We need a way to use Kerberos with a cached credential which by definition is an empty username/password in the `credential`. In this case the `username` must have a valid so we need to use the `SshClientSettings.UserName` value so it's a problem that needs to be solved now. I think the approach that I have in this PR right now is a decent approach that aligns with the behaviour of the normal `ssh` binary where the SSH username is the explicitly provided value in the connection string (or `-o UserName=...` with a fallback to the current user) and it will use the cached Kerberos credential for Kerb auth which can be a separate user.

The only scenario not covered here is if you want to use a different SSH username when providing an explicit `credential`. This could be solved in the future using a separate string or bool option that says to favour the `SshClientSettings.UserName` over the explicitly provided `credential` but I don't think we need to solve that use case right now.

tmds: Thanks for the clear explanation!

> Another option is to always treat the SshClientSettings.UserName as the SSH username provided and only use the explicit credential for the Kerberos auth exchange. Thinking about this more this might actually be a better option than what I have today as it supports all the scenarios I can think off.

:+1: I was thinking the same thing as I read through your text.

jborean93: Thanks for asking the right questions, always good to go through the details and figure out better ways of doing things!

tmds: Totally agree!

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// Credentials can only be delegated if the Kerberos ticket retrieved from the KDC is marked as forwardable.
+    /// Windows hosts will always retrieve a forwardable ticket but non-Windows hosts may not. When using an explicit
+    /// credential, make sure that 'forwardable = true' is set in the krb5.conf file so that .NET will request a
+    /// forwardable ticket required for delegation. When using a cached credential, make sure that when the ticket was
+    /// retrieved it was retrieved with the forwardable flag. If the ticket is not forwardable, the authentication will
+    /// still work but the ticket will not be delegated.
+    /// </remarks>
+    /// <param name="credential">The credentials to use for the Kerberos authentication exchange. Set to null to use the cached credential.</param>
+    /// <param name="delegateCredential">Request delegation on the Kerberos context.</param>
+    /// <param name="serviceName">Override the service principal name (SPN), default uses the <c>host/{connection.HostName}.</c></param>
```
tmds, line 61: I read through this and made some edits. ptal and include the changes that make sense.

```
/// <summary>
/// Create a credential for Kerberos authentication.
/// </summary>
/// <remarks>
/// The credential specified is used for the Kerberos authentication process. This can either be the same or
/// different from the username specified through <c>SshClientSettings.UserName</c>. The client settings username
/// is the target login user the SSH service is meant to run as, whereas the credential is the Kerberos
/// principal used for authentication. The rules for how a Kerberos principal maps to the target user is defined by
/// the SSH service itself. For example on Windows the username should be the same but on Linux the mapping can be
/// done through a <c>.k5login</c> file in the target user's home directory.
///
/// If the credential is <c>null<c>, the Kerberos authentication will be done using a cached ticket.
/// For Windows, this is the current thread's identity (typically logon user) will be used.
/// For Unix/Linux, this will use the Kerberos credential cache principal, which may  be managed using the
/// <c>kinit</c> command. If there is no available cache credential, the authentication will fail.
///
/// Credentials can only be delegated if the Kerberos ticket retrieved from the KDC is marked as forwardable.
/// Windows hosts will always retrieve a forwardable ticket but non-Windows hosts may not. When using an explicit
/// credential, make sure that 'forwardable = true' is set in the krb5.conf file so that .NET will request a
/// forwardable ticket required for delegation. When using a cached ticket, make sure that when the ticket was
/// retrieved it was retrieved with the forwardable flag. If the ticket is not forwardable, the authentication will
/// still work but the ticket will not be delegated.
/// </remarks>
/// <param name="credential">The credentials to use for the Kerberos authentication exchange. Set to null to use a cached ticket.</param>
/// <param name="delegateCredential">Allows the SSH server to impersonate the user on remote systems.</param>
/// <param name="serviceName">Override the service principal name (SPN), default uses the <c>host/<SshClientSettings.Host>.</c></param>
```

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,294 @@
...
+    public KerberosCredential(NetworkCredential? credential = null, bool delegateCredential = false, string? serviceName = null)
+    {
```
tmds, line 63: When a user provides us with a non-empty UserName in the credential we should throw if the Password is empty.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,294 @@
...
+        // RFC uses hostbased SPN format "service@host" but Windows SSPI needs the service/host format.
+        // The latter works on both SSPI and GSSAPI so we use that as the default.
+        string spn = string.IsNullOrEmpty(_serviceName) ? $"host/{connectionInfo.Host}" : _serviceName;
+
+        // The SSH messages must have a username value which maps to the target user we want to login as. We use the
+        // client supplied username as the target user. The Kerberos principal credential is only used for the
+        // authentication stage that happens next.
+        string userName = settings.UserName;
+        logger.AuthenticationMethodGssapiWithMic(userName, spn, _delegateCredential);
```
tmds, line 79: Let's include the credential username too in the logged message.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,294 @@
...
+        writer.WriteString("ssh-connection");
+        writer.WriteString(AUTH_METHOD);
+
+        // The MIC data does not include the header, so skip the first 5 bytes.
+        ReadOnlySequence<byte> micData = packet.Move().AsReadOnlySequence().Slice(5);
+
+        var signatureWriter = new ArrayBufferWriter<byte>();
+
+#if NET8_0
+        GetMICMethod(
```
tmds, line 242: If this method isn't found, calling it will throw `MissingMethodException`.
We should add a catch for that in `TryAuthenticate`, and return `false`.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,294 @@
...
+    /// <param name="delegateCredential">Request delegation on the Kerberos context.</param>
+    /// <param name="serviceName">Override the service principal name (SPN), default uses the <c>host/{connection.HostName}.</c></param>
+    public KerberosCredential(NetworkCredential? credential = null, bool delegateCredential = false, string? serviceName = null)
+    {
+        _kerberosCredential = credential ?? CredentialCache.DefaultNetworkCredentials;
+        _delegateCredential = delegateCredential;
+        _serviceName = serviceName;
+    }
+
+    internal async Task<bool> TryAuthenticate(SshConnection connection, ILogger logger, SshClientSettings settings, SshConnectionInfo connectionInfo, CancellationToken ct)
```
tmds, line 69: When required libraries are missing (`krb5`) this method shouldn't throw and return `false`.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Makes sense.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Looks good, only made a tiny tweak to the `delegateCredential` to change `impersonate` to `delegate` because impersonate is a concept used on Windows for local operations whereas delegation is a step beyond that.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Makes sense

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+            }
+
+            ReadOnlySequence<byte> oidResponse = GetGssapiOidResponse(response);
+            if (KRB5_OID.AsSpan().SequenceEqual(oidResponse.IsSingleSegment ? oidResponse.FirstSpan : oidResponse.ToArray()))
+            {
+                return true;
+            }
+            else
+            {
+                var reader = new AsnReader(oidResponse.ToArray(), AsnEncodingRules.DER);
```
tmds, line 168: If KRB5_OID is already DER encoder, why do we need this branch?

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+    {
+        byte[]? outToken;
+        NegotiateAuthenticationStatusCode statusCode;
+        try
+        {
+            outToken = authContext.GetOutgoingBlob(Array.Empty<byte>(), out statusCode);
+        }
+        catch (InvalidOperationException)
+        {
+            logger.AuthenticationKerberosFailed("Kerberos library not available");
```
tmds, line 186: I think we might get `NegotiateAuthenticationStatusCode.Unsupported` when the Kerberos library is not available.

Did you try and see an `InvalidOperationException`?

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+
+        while (outToken is not null)
+        {
+            {
+                using var userAuthMsg = CreateGssapiTokenMessage(connection.SequencePool, outToken);
+                await connection.SendPacketAsync(userAuthMsg.Move(), ct).ConfigureAwait(false);
+            }
+
+            // If the context is complete we don't expect a response.
+            if (statusCode == NegotiateAuthenticationStatusCode.Completed)
```
tmds, line 198: Can we change this condition to one that checks the status for continuing the loop?
Perhaps: `if (statusCode != NegotiateAuthenticationStatusCode.ContinueNeeded`)?

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+    private static Packet CreateGssapiMicData(SequencePool sequencePool, ReadOnlySpan<byte> sessionId, string userName, NegotiateAuthentication authContext)
+    {
+        /*
+            string    session identifier
+            byte      SSH_MSG_USERAUTH_REQUEST
+            string    user name
+            string    service
+            string    "gssapi-with-mic"
+        */
+        using var packet = sequencePool.RentPacket();
```
tmds, line 282: A `Packet` is a small wrapper around a `Sequence` which is used to deal with 'packet' stuff.
This data isn't a packet, so you could rent a `Sequence` directly and use a `SequenceWriter` to write to it.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+        using var authContext = new NegotiateAuthentication(negotiateOptions);
+        bool isAuthSuccess = await TryStageAuthentication(connection, logger, authContext, ct).ConfigureAwait(false);
+        if (!isAuthSuccess)
+        {
+            return false;
+        }
+
+        try
+        {
+            using var message = authContext.IsSigned
```
tmds, line 117: What determines whether `IsSigned` is `true`/false`?

Can you add some comment that explains why there are two cases here?

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+                                        userName, KRB5_OID);
+            await connection.SendPacketAsync(userAuthMsg.Move(), ct).ConfigureAwait(false);
+        }
+
+        while (true)
+        {
+            using Packet response = await connection.ReceivePacketAsync(ct).ConfigureAwait(false);
+            MessageId? mid = response.MessageId;
+
+            if (mid == MessageId.SSH_MSG_USERAUTH_BANNER)
```
tmds, line 144: +1

There are probably other cases where this message needs to be filtered out. I created https://github.com/tmds/Tmds.Ssh/issues/192 to look into it further.

README.md
```diff
@@ -373,6 +373,11 @@ class PasswordCredential : Credential
   PasswordCredential(string password);
   PasswordCredential(Func<string?> prompt);
 }
+class GssapiWithMicCredential : Credential
```
tmds, line 376: This needs an update.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+    /// Windows hosts will always retrieve a forwardable ticket but non-Windows hosts may not. When using an explicit
+    /// credential, make sure that 'forwardable = true' is set in the krb5.conf file so that .NET will request a
+    /// forwardable ticket required for delegation. When using a cached ticket, make sure that when the ticket was
+    /// retrieved it was retrieved with the forwardable flag. If the ticket is not forwardable, the authentication will
+    /// still work but the ticket will not be delegated.
+    /// </remarks>
+    /// <param name="credential">The credentials to use for the Kerberos authentication exchange. Set to null to use a cached ticket.</param>
+    /// <param name="delegateCredential">Allows the SSH server to delegate the user on remote systems.</param>
+    /// <param name="serviceName">Override the service principal name (SPN), default uses the <c>host/<SshClientSettings.Host>.</c></param>
+    public KerberosCredential(NetworkCredential? credential = null, bool delegateCredential = false, string? serviceName = null)
```
tmds, line 62: Can you consider the names and the order of these arguments, and feel free to change what you want if you think it improves the usability.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,339 @@
...
+
+namespace Tmds.Ssh;
+
+public sealed class KerberosCredential : Credential
+{
+    // gssapi-with-mic is defined in https://datatracker.ietf.org/doc/html/rfc4462
+    private const string AUTH_METHOD = "gssapi-with-mic";
+
+    // Kerberos - 1.2.840.113554.1.2.2
+    private static readonly byte[] KRB5_OID = [ 0x06, 0x09, 0x2A, 0x86, 0x48, 0x86, 0xF7, 0x12, 0x01, 0x02, 0x02 ];
```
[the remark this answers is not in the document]
jborean93: Yep, will do.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: It's to log the response OID from the server when it doesn't match what we sent. I'm not sure if this could happen but it's to verify the server is responding the way we would expect and to provide a way to see what happens in the failure scenarios through the logs.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Looks like you are right. I was looking at https://github.com/dotnet/runtime/blob/6e88cf8c62a8100c36eef4bcc26a3d830b60ff35/src/libraries/System.Net.Security/src/System/Net/NegotiateAuthenticationPal.Unsupported.cs#L52-L56 and saw the `InvalidOperationException` but didn't realise `GetOutgoingBlock` was doing something different.

I'll update the code to handle this.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Sure I'll have a look, I think this new logic should work.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: It's based on the SSPI/GSSAPI flags in the exchange. We request both signing/encryption https://github.com/jborean93/Tmds.Ssh/blob/ed9e230bb60e3fca7deac7ba92caf2354c58c541/src/Tmds.Ssh/KerberosCredential.cs#L100 but the server can technically not honour this and not reply with the signing/encryption flags. While all implementations I know off will always sign for Kerberos I'm just following the spec to have both scenarios here. I'll add a comment.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: Sounds good, I came across it when manually testing against a banner. It seems like only the first message in the exchange contains the banner. Having a generic way to handle this would be great but for now I'm just ignoring it like the code in `UserAuthentication.cs`.

src/Tmds.Ssh/KerberosCredential.cs
jborean93: We talked about the arg names earlier and I settled on these because they align with the Kerberos/GSSAPI nomenclature. I think the ordering makes sense here but is there something in particular you are concerned about.

tmds: No concern from my side. I'm double checking because this is the public API and I find it important we have good usability.

src/Tmds.Ssh/KerberosCredential.cs
tmds: Ah yes, this is for logging.

I'd rather avoid the `AsnReader` here (to avoid unexpected issues on an unlikely branch), can you change it to `Convert.ToBase64String` the `oidResponse`?

jborean93: Sure thing.

tmds:
@jborean93 thanks a lot for contributing this very useful feature!

tmds:
This is part of 0.4.0 which was just uploaded to nuget.org.

src/Tmds.Ssh/KerberosCredential.cs
```diff
@@ -0,0 +1,327 @@
...
+
+        var negotiateOptions = new NegotiateAuthenticationClientOptions()
+        {
+            AllowedImpersonationLevel = _delegateCredential ? TokenImpersonationLevel.Delegation : TokenImpersonationLevel.Impersonation,
+            Credential = _kerberosCredential,
+            Package = "Kerberos",
+            // While only Sign is needed we need to set EncryptAndSign for
+            // Windows client support. Sign only will pass in SECQOP_WRAP_NO_ENCRYPT
+            // to MakeSignature which fails.
+            // https://github.com/dotnet/runtime/issues/103461
```
tmds, line 98: hi @jborean93, the issue you referenced here is closed. Is there some change we can/should make for .NET9+?

jborean93: With the fix we can change it to `RequiredProtectionLevel = ProtectionLevel.Sign;`. but I don't think it ended up being backported to net9 so the fix is only present in net10. The PR which fixed it https://github.com/dotnet/runtime/pull/105605 was merged after net9 was created and I don't see any reference to a backport.

tmds: Thanks for the input! I assumed it was fixed in .NET 9 based on the issue milestone.
We can add the change under an `#if NET10_0_OR_GREATER`.

PR #196, opened 2024-06-24
DefaultCredentials: add KerberosCredential.

tmds:
@jborean93 ptal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ class SftpFile : Stream
 }
 class SshClientSettings
 {
-  static IReadOnlyList<Credential> DefaultCredentials { get; } // = [ "~/.ssh/id_rsa" ]
+  static IReadOnlyList<Credential> DefaultCredentials { get; } // = [ PrivateKeyCredential("~/.ssh/id_rsa"), KerberosCredential() ]
 
   SshClientSettings();
   SshClientSettings(string destination);
--- a/src/Tmds.Ssh/SshClientSettings.cs
+++ b/src/Tmds.Ssh/SshClientSettings.cs
@@ -138,7 +138,8 @@ public sealed partial class SshClientSettings
         [
             // new PrivateKeyCredential(Path.Combine(home, ".ssh", "id_ed25519")),
             // new PrivateKeyCredential(Path.Combine(home, ".ssh", "id_ecdsa")),
-            new PrivateKeyCredential(Path.Combine(home, ".ssh", "id_rsa"))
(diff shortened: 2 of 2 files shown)


tmds:
@jborean93 I assume it makes sense to include this in the `DefaultCredentials`?

jborean93:
I personally think so, my only concern is I missed a spot where an exception could be thrown. While I think we covered them I cannot say for sure I did. One other point to consider is that OpenSSH doesn't include it by default, you either need to specify `-k` or add it in your ssh config file.

tmds:
I'm not worried a potential exception. If it's there, we'll fix it.

Might there some information we're leaking to the remote SSH server about the user, like its Kerberos username?

jborean93:
> Might there some information we're leaking to the remote SSH server about the user, like its Kerberos username?

Certainly if there is a cached credential you could be leaking the username but the sensitive information are all part of the encrypted value that only the intended target would be able to decrypt.

tmds:
> only the intended target would be able to decrypt.

Paraphrasing to see if I understand correctly: if the target isn't part of the "domain", it won't be able to decrypt the username?

tmds:
> Paraphrasing to see if I understand correctly: if the target isn't part of the "domain", it won't be able to decrypt the username (or might even never see the username)?

To answer this question, from the Wikipedia Kerberos page: _Client-to-server ticket, encrypted using service's Secret key_.

@jborean93 thanks for chiming in! I'm going to include this.

jborean93:
> Paraphrasing to see if I understand correctly: if the target isn't part of the "domain", it won't be able to decrypt the username (or might even never see the username)?

Just to add further information the ticket won’t even be generated if the target server isn’t valid/present in the KDC database.

tmds:
@jborean93 do the methods we call on the `NegotiateAuthentication` cause network packets to be exchanged with the Kerberos servers? If so, we should probably do some work to make these methods run asynchronously in order to respect the `CancellationToken`?

jborean93:
Retrieving the service ticket does a call to the KDC in the backend to the call to get the outgoing blob may block. As this happens in the underlying C library I don't believe .NET exposes an async API for this. I believe it would need to run in `Task.Run` to support async. The other calls to `NegotiateAuthentication` should not require any network IO.

tmds:
Thanks for confirming. Yes, it will have to be `Task.Run`.

I will look into making those changes and ask you for review. If you would want to pick it up, let me know.

PR #197, opened 2024-06-25
Make auth cancellable during NegotiateAuthentication.

tmds:
@jborean93 ptal.


jborean93 (approved):
At a look the changes look good to m,  I'm not too familiar this the cancellation model you have here but from a read through seems ok to me.

PR #198, opened 2024-06-28
add Tmd.Ssh.sln solution file and added all projects to .sln file

almesalati:


tmds:
@almesalati do you find it helpful for the repo to include an sln file?

I won't be using this myself, so it may bitrot.

almesalati:
The solution file helps when working with visual studio, and i think it will be helpful 
Thanks for your time, @tmds

tmds:
Thanks @almesalati!

Issue #203, opened 2024-07-25
ECDSA key support


jborean93:
Issue to track support for ECDSA keys which should be possible with the builtin [ECDsa class](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.ecdsa?view=net-8.0).

tmds:
This is part of 0.5.0 which was just published to nuget.org.

PR #206, opened 2024-07-25
Implement ECDSA private key support

jborean93:
Add the ability to use ECDSA private keys for user authentication. These keys support the NIST P_256, P_384, and P_521 curves that can be generated by ssh-keygen.

Fixes: https://github.com/tmds/Tmds.Ssh/issues/203


src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
```diff
@@ -224,45 +224,127 @@ byte	3
...
+                };
+                if (parameters.D.Length < q.X!.Length)
+                {
+                    // ToByteArray will shorten the value to the minimum size,
+                    // we need to prepend 0 to the array to make it the
+                    // size.
+                    byte[] originalD = parameters.D;
+                    byte[] expandedD = new byte[q.X.Length];
+                    Buffer.BlockCopy(originalD, 0, expandedD, expandedD.Length - originalD.Length, originalD.Length);
+                    parameters.D = expandedD;
```
jborean93, line 336: I'm not 100% sure on this one. I don't know if this is a side effective of the P521 curve or just the ECDSA cert I generated but it was always 1 byte short. I couldn't just use `ReadStringAsByteArray()` as that also returned the shorter byte array.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
```diff
@@ -224,45 +224,127 @@ byte	3
...
                 }
-                catch (Exception ex)
+
+                ECParameters parameters = new()
                 {
-                    error = new FormatException($"The data can not be parsed into an RSA key.", ex);
-                    rsa?.Dispose();
-                    return false;
+                    Curve = curve,
+                    D = d.ToByteArray(true, true),
```
jborean93, line 325: While `ReadMPInt()` reads the raw bytes as a signed value, I found I had to specify the `isUnsigned: true` here or else it would include the signed bit and be too large for `ECParameters`.

README.md
```diff
@@ -400,7 +400,7 @@ This section lists the currently supported algorithms. If you would like support
 
 Supported private key formats:
 - RSA in `RSA PRIVATE KEY`
-- RSA in `OPENSSH PRIVATE KEY` (`openssh-key-v1`)
+- RSA, ECDSA in `OPENSSH PRIVATE KEY` (`openssh-key-v1`)
```
jborean93, line 403: I don't know if ECDSA keys in SSH can actually use other curves than the 3 emitted by `ssh-keygen` and implemented in this PR. I can explicitly call them out if you wish or just leave it generic like it is here.

tmds:
@jborean93, thanks for implementing this! I'll review it this week.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
```diff
@@ -224,45 +224,127 @@ byte	3
...
+            try
+            {
+                byte[] dBytes = d.ToByteArray(true, true);
+                if (dBytes.Length < q.X!.Length)
                 {
-                    rsa.ImportParameters(parameters);
-                    privateKey = new RsaPrivateKey(rsa);
-                    error = null;
-                    return true;
+                    Array.Resize(ref dBytes, q.X.Length);
```
tmds, line 319: How did you come to add a resize here?

I think if we need to resize, we should be prepending with zeros (for big endian), otherwise we'll change the numeric value.

tmds: I'm looking, but I don't see where `dBytes` is used.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
tmds: Can you check if `BigInteger d` is a positive or a negative number.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
tmds: mpint is defined in https://datatracker.ietf.org/doc/html/rfc4251#section-5. We need to read it as a signed (that is: 2-complement) value.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
tmds: From https://github.com/tmds/Tmds.Ssh/pull/206/files#r1693915325, I assume something expects parameters.D.Length and q.X!.Length to be the same.

Then it makes sense to have to pre-pend D with zeros when it is a big-endian unsigned value.
That way we meet the length expectation without changing the numeric value of D.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
jborean93: Sorry this was leftover code I was using during testing I forgot to remove.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
jborean93: > I assume something expects parameters.D.Length and q.X!.Length to be the same.

Yep importing the parameters in the `ECDsa` object fails if the length of `D` does not match the curve lengths. When the value is signed it contains an extra byte and I found in the P521 example it didn't contain enough bytes so needed to be left padded with 0s.

> Can you check if BigInteger d is a positive or a negative number.

Using the 3 different curves set up in the test private key we have the following values

P256 - 8.35053129e+76 - Sign 1
00B89E550B526C3BEF46B130997C32560452C5011EFDFFB48820CE38D91F6AD2FD
00B89E550B526C3BEF46B130997C32560452C5011EFDFFB48820CE38D91F6AD2FD
B89E550B526C3BEF46B130997C32560452C5011EFDFFB48820CE38D91F6AD2FD


P384 - 2.45161837e+115 - Sign 1
009F48EC8F7EB1705D63DB6AD3F26A6B37E340E7944F30F7EC2CDCC65337AED6472F95721136013C20659CC4558C0002FF
009F48EC8F7EB1705D63DB6AD3F26A6B37E340E7944F30F7EC2CDCC65337AED6472F95721136013C20659CC4558C0002FF
9F48EC8F7EB1705D63DB6AD3F26A6B37E340E7944F30F7EC2CDCC65337AED6472F95721136013C20659CC4558C0002FF


P521 - 1.00062567e+155 - Sign 1
077687A370EF8BD1C78C891537CA857AFDB32FC95D56CD5CD501F527F399CC17C136D8C8CF0FB630E4C62DEAFF61CCDFC4F6C0272AC92995FB603214DF86CC2EE8
077687A370EF8BD1C78C891537CA857AFDB32FC95D56CD5CD501F527F399CC17C136D8C8CF0FB630E4C62DEAFF61CCDFC4F6C0272AC92995FB603214DF86CC2EE8
077687A370EF8BD1C78C891537CA857AFDB32FC95D56CD5CD501F527F399CC17C136D8C8CF0FB630E4C62DEAFF61CCDFC4F6C0272AC92995FB603214DF86CC2EE8

The first value is the raw MPInt bytes after the length, the second is `ToByteArray` as a signed value and the latter is as a unsigned value.

It looks like I should be doing `ToByteArray(isUnsigned: false, isBigEndian: true)` but I need to handle these two both the extra 0 padding that makes the length too long and the case when there isn't enough padding to begin with.

I'll do some testing and generate some more ECDSA certs to see if there are other scenarios I'm not thinking off here.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
jborean93: I've added a new overload `ReadMPIntAsBytes()` that returns the `ReadOnlySequence<byte>` which avoiding having to build and then get the bytes of a `BigInteger` value altogether.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
jborean93: The latest commits change the logic here to just read the raw bytes and handle when D is the required size, contains the leading 0 byte sign char, needs to be left padded. I've also updated the test keys where P256 is the required size, P384 contains the 0 byte sign char, and P521 needs to be left padded with 0s.

README.md
tmds: This is good.

A bit lower, you can add the specific algorithms under "Supported client key algorithms:".

src/Tmds.Ssh/ECDsaPrivateKey.cs
```diff
@@ -0,0 +1,76 @@
...
+sealed class ECDsaPrivateKey : PrivateKey
+{
+    private readonly ECDsa _ecdsa;
+    private readonly Name _ecIdentifier;
+    private readonly Name _curveName;
+    private readonly HashAlgorithmName _allowedAlgorithm;
+
+    public ECDsaPrivateKey(ECDsa ecdsa, Name identifier, Name curveName, HashAlgorithmName allowedAlgorithm) :
+        base([identifier])
+    {
```
tmds, line 21: Can you rename:
`identifier` -> `algorithm`
`_ecIdentifier` -> `_algorithm`
`allowedAlgorithm` -> `hashAlgorithm`
`_allowedAlgorithm` -> `_hashAlgorithm`.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
```diff
@@ -224,45 +224,132 @@ byte	3
...
+                curve = ECCurve.NamedCurves.nistP384;
+            }
+            else if (curveName == AlgorithmNames.Nistp521)
+            {
+                allowedHashAlgo = HashAlgorithmName.SHA512;
+                curve = ECCurve.NamedCurves.nistP521;
+            }
+            else
+            {
+                error = new FormatException($"ECDSA curve '{curveName}' is unsupported.");
```
tmds, line 306: We should use `NotSupportedException` instead of `FormatException` here.

test/Tmds.Ssh.Tests/PrivateKeyCredentialTests.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading.Tasks;
+
+namespace Tmds.Ssh.Tests;
+
+[Collection(nameof(SshServerCollection))]
+public class PrivateKeyCredentialTests
+{
+    private readonly SshServer _sshServer;
+
+    public PrivateKeyCredentialTests(SshServer sshServer) : base()
```
tmds, line 12: nit: drop the `: base()`.

src/Tmds.Ssh/UserAuthentication.PublicKeyAuth.cs
tmds: I pushed a commit that moved the padding logic into the SequenceReader.

src/Tmds.Ssh/ECDsaPrivateKey.cs
jborean93: Sure

README.md
jborean93: Ah I missed that section, have added it there.

tmds:
Thanks @jborean93!

jborean93:
Thanks for the speedy review and help with the MPInt handling! I'll rebase my encrypted key PR with the changes merged here, no rush on the review there though.

PR #207, opened 2024-07-26
Add support for encrypted private keys.

jborean93:
Adds support for using encrypted private keys with user authentication. This implementation support private keys encrypted with the PKCS#1 and OpenSSH PEM formats. The AES and ChaCha20-Poly1305 ciphers allowed by each format has been implemented in this change.

Fixes: https://github.com/tmds/Tmds.Ssh/issues/202


jborean93:
I tried my best to clean up the Bcrypt KDF implementation from SSH.NET but I'm sure there are probably further optimisations that could be implemented. The good news is it works with some test vectors I found and the tests when they run.

src/Tmds.Ssh/ChaCha20Decrypter.cs
```diff
@@ -0,0 +1,24 @@
...
+        // https://github.com/openssh/openssh-portable/blob/master/PROTOCOL.chacha20poly1305
+        Span<byte> nonce = stackalloc byte[12];
+        Span<byte> tag = stackalloc byte[16];
+        tag[0] = 1;  // K_2 sets the counter to 1
+        byte[] decData = new byte[data.Length];
+
+        using var chacha = new ChaCha20Poly1305(key);
+        chacha.Encrypt(nonce, data, decData, tag);
+
+        return decData;
```
jborean93, line 22: This works in the tests but I don't have a great understanding of this protocol so I'm unsure if this is ok. I mostly just saw how the Python library asyncssh did the key decryption and this was it. They did have a verify_tag call which derived the Poly1305 key but considering we are verifying the checkint on the decrypted data I'm unsure if it is worth doing the same.

src/Tmds.Ssh/EncryptionAlgorithm.cs
```diff
@@ -3,18 +3,21 @@
...
 
 namespace Tmds.Ssh;
 
 sealed class EncryptionAlgorithm
 {
+    private delegate byte[] DecryptDelegate(ReadOnlySpan<byte> key, Span<byte> iv, ReadOnlySpan<byte> data, ReadOnlySpan<byte> tag);
+
     private readonly Func<EncryptionAlgorithm, byte[], byte[], HMacAlgorithm?, byte[], IPacketEncoder> _createPacketEncoder;
     private readonly Func<EncryptionAlgorithm, SequencePool, byte[], byte[], HMacAlgorithm?, byte[], IPacketDecoder> _createPacketDecoder;
+    private readonly DecryptDelegate _decryptData;
```
tmds, line 15: This class is meant as a factory for packet decoders and packet encoders. You need a factory for encryption algorithms. Let's keep them separate by renaming the existing `EncryptionAlgorithms` to `PackagingAlgorithms` and then you can use `EncryptionAlgorithms` for the new factory.

src/Tmds.Ssh/ChaCha20Decrypter.cs
tmds: See https://github.com/dotnet/runtime/issues/69130#issuecomment-2133268304 for some specifics about OpenSSL chacha20poly1305 vs .NET ChaCha20Poly1305.

If we're unsure, I'm fine to leave out chacha20poly1305 for now. I assume it is not used much for encrypting private keys.

src/Tmds.Ssh/PrivateKeyCredential.OpenSsh.cs
```diff
@@ -0,0 +1,272 @@
...
+using System.Buffers;
+using System.Diagnostics.CodeAnalysis;
+using System.Numerics;
+using System.Security.Cryptography;
+
+namespace Tmds.Ssh;
+
+partial class PrivateKeyCredential
+{
+    internal static bool TryParseOpenSshKey(
```
tmds, line 14: Can you move this to `PublicKeyAuth` or under a new helper class, like `PrivateKeyParser`.

src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -2,15 +2,21 @@
...
+public sealed partial class PrivateKeyCredential : Credential
 {
     internal string FilePath { get; }
+    internal byte[] Passphrase { get; }
 
-    public PrivateKeyCredential(string path)
+    public PrivateKeyCredential(string path) : this(path, null)
+    { }
+
+    public PrivateKeyCredential(string path, string? passphrase)
```
tmds, line 17: Can we use 'password', like `X509Certificate2` ctor?

src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -2,15 +2,21 @@
...
     {
         FilePath = path ?? throw new ArgumentNullException(nameof(path));
+        Passphrase = string.IsNullOrEmpty(passphrase) ? Array.Empty<byte>() : Encoding.UTF8.GetBytes(passphrase);
```
tmds, line 20: Let's store this as a string (same type as passed to the constructor).

README.md
```diff
@@ -367,6 +367,7 @@ abstract class Credential
 class PrivateKeyCredential : Credential
 {
   PrivateKeyCredential(string path);
+  PrivateKeyCredential(string path, string? passphrase)
```
tmds, line 370: You can merge this with the existing overload, and initialize `= null`.

I think it's ok to do a minor breaking change like this at this point to reduce the number of overloads a user needs to work with.

tmds:
@jborean93 thanks for working on this! I've added some high level comments about code organization. I'll try to do a more in depth review in the coming week.

src/Tmds.Ssh/ChaCha20Decrypter.cs
jborean93: Makes sense, I think I'll just open an issue with the details from this PR so it can be revisited later on.

src/Tmds.Ssh/PrivateKeyCredential.cs
jborean93: I can change it if you like. The reason why I went with `passphrase` is that's the wording OpenSSH uses for encrypted private keys.

src/Tmds.Ssh/PrivateKeyCredential.OpenSsh.cs
jborean93: I went with the latter.

src/Tmds.Ssh/EncryptionAlgorithm.cs
jborean93: Will do, I can't say I was happy with adding more stuff on top of this class as it didn't seem like it fit very well. I'll keep them separate as you've recommended.

src/Tmds.Ssh/PrivateKeyCredential.cs
jborean93: I chose a `byte[]` so that the parser code didn't need to do the encoding itself. It could also be possible to use a passphrase of non-UTF8 bytes if an overload was added to accept `byte[]` alongside a string.

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: I prefer `password` because that is what .NET BCL API use.

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: I don't think we should account for non encode-able passwords.

The conversion to bytes is cheap enough to move it to the parser, and have the `Credential` class as dumb as possible. The credential classes are simple data holders that expose the constructor arguments as internal properties.

src/Tmds.Ssh/AesCtr.cs
```diff
@@ -0,0 +1,47 @@
...
+// See file LICENSE for full license details.
+
+using System;
+using System.Security.Cryptography;
+
+namespace Tmds.Ssh;
+
+static class AesCtr
+{
+    public static byte[] DecryptCtr(ReadOnlySpan<byte> key, Span<byte> counter, ReadOnlySpan<byte> data)
```
tmds, line 11: Let's accept the argument to decode to as a `Span` so the caller can manage the memory.

```
public static byte[] DecryptCtr(ReadOnlySpan<byte> key, Span<byte> counter, ReadOnlySpan<byte> data)
```
to
```
public static void DecryptCtr(ReadOnlySpan<byte> key, Span<byte> counter, ReadOnlySpan<byte> data` -> `ReadOnlySpan<byte> ciphertext, Span<byte> plaintext)
```

src/Tmds.Ssh/PackingAlgorithm.cs
```diff
@@ -3,18 +3,17 @@
 
 using System;
 using System.Collections.Generic;
-using System.Security.Cryptography;
 
 namespace Tmds.Ssh;
 
-sealed class EncryptionAlgorithm
+sealed class PackingAlgorithms
```
tmds, line 9: Let's keep the old name (since we're using a different name for the openssl key ciphers).

src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs
```diff
@@ -0,0 +1,279 @@
...
+        if (!OpenSshKeyCipher.TryGetCipher(cipher, out var keyCipher))
+        {
+            error = new NotSupportedException($"Unsupported Cipher: '{cipher}'.");
+            return false;
+        }
+
+        try
+        {
+            byte[] derivedKey = new byte[keyCipher.KeyLength + keyCipher.IVLength];
+            new BCrypt().Pbkdf(
```
tmds, line 152: Can you add a static method to `BCrypt` that we call here, and name it something like:
`DeriveKeyFromPassword`. The static method can call what you are calling here.

src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -2,15 +2,18 @@
...
 public sealed class PrivateKeyCredential : Credential
 {
     internal string FilePath { get; }
+    internal byte[] Passphrase { get; }
 
-    public PrivateKeyCredential(string path)
+    public PrivateKeyCredential(string path, string? passphrase = null)
     {
         FilePath = path ?? throw new ArgumentNullException(nameof(path));
+        Passphrase = string.IsNullOrEmpty(passphrase) ? Array.Empty<byte>() : Encoding.UTF8.GetBytes(passphrase);
```
tmds, line 17: Can we treat the empty non-null string as a password that is empty (instead of it to mean the same thing as `null`)?

tmds: ... because though it isn't a good idea to use an empty string as a password, I assume it is possible.

jborean93: It may be possible with `openssl` to generate a PKCS#1 key but `ssh-keygen` treats `-N ''` to be no password. It probably is possible to build the encryption key from an empty string manually but not using the normal tools. Because of this I thought treating an empty string the same as `null`/no password. It was also done because the underlying method to decrypt the code was using a `ReadOnlySpan<byte>` so was easier to treat an empty span as no password set.

src/Tmds.Ssh/AlgorithmNames.cs
```diff
@@ -35,11 +35,35 @@ static class AlgorithmNames // TODO: rename to KnownNames
...
+    private static readonly byte[] Pkcs1Aes128CbcBytes = "AES-128-CBC"u8.ToArray();
+    public static Name Pkcs1Aes128Cbc => new Name(Pkcs1Aes128CbcBytes);
+    private static readonly byte[] Pkcs1Aes192CbcBytes = "AES-192-CBC"u8.ToArray();
+    public static Name Pkcs1Aes192Cbc => new Name(Pkcs1Aes192CbcBytes);
+    private static readonly byte[] Pkcs1Aes256CbcBytes = "AES-256-CBC"u8.ToArray();
+    public static Name Pkcs1Aes256Cbc => new Name(Pkcs1Aes256CbcBytes);
+
+    // KDF algorithms:
+    private static readonly byte[] BcryptBytes = "bcrypt"u8.ToArray();
+    public static Name Bcrypt => new Name(BcryptBytes);
```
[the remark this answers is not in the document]
jborean93: Fixed to use `BCrypt`.

src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs
jborean93: I made this a static method so kept the `Pbkdf` name, let me know if you prefer `DeriveKeyFromPassword`.

tmds: Yes, I prefer a more descriptive name for the method.

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: Ok, let's keep it as is.

src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs
jborean93: No worries, have updated the code.

src/Tmds.Ssh/PrivateKeyParser.RsaPkcs1.cs
```diff
@@ -0,0 +1,147 @@
...
+using System.Security.Cryptography;
+
+namespace Tmds.Ssh;
+
+partial class PrivateKeyParser
+{
+    /// <summary>
+    /// Parses an RSA PKCS#1 PEM formatted key. This is a legacy format used
+    /// by older ssh-keygen and openssl versions for RSA based keys.
+    /// </summary>
```
tmds, line 16: Do you think it's important to support this older format?

src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs
```diff
@@ -0,0 +1,279 @@
...
+            byte[] derivedKey = new byte[keyCipher.KeyLength + keyCipher.IVLength];
+            BCrypt.DeriveKeyFromPassword(
+                password,
+                kdfSalt.IsSingleSegment ? kdfSalt.FirstSpan : kdfSalt.ToArray(),
+                (int)rounds,
+                derivedKey);
+
+            ReadOnlySequence<byte> encryptedKey = reader.ReadStringAsBytes();
+            ReadOnlySequence<byte> tag = default;
+            if (keyCipher.IsAuthenticated && keyCipher.TagLength > 0)
```
tmds, line 160: keyCipher.IsAuthenticated and TagLength > 0 are the same thing. We can drop this `if` and unconditionally execution its body.

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: I'd like to come back on this. I think the .NET semantics should be null for no password and not null for a password.

Next to the Span, we can add a bool that tracks if a password was provided.

jborean93: Sure we can do that

src/Tmds.Ssh/PrivateKeyParser.RsaPkcs1.cs
jborean93: Probably less important as time goes on but it was the default format used by ssh-keygen and openssl for a long time and there’s probably still people running on older hosts. The implementation is quite simple as well so from a maintenance perspective it’s not too bad. The only reason I can see for not implementing it would be to force people to use a newer and more secure format but at that point you may ad well not support the key format altogether (even as plaintext).

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: Can you also add a `Func<string?> prompt)` overload, like we have for `PasswordCredential`?

src/Tmds.Ssh/PrivateKeyParser.RsaPkcs1.cs
tmds: Do you have an idea when `ssh-keygen` was using this format (as the default)?

tmds: I'm fine including this, if you think it is still relevant.

jborean93: Looks like it was OpenSSH 7.8 (2018-08-24) https://www.openssh.com/txt/release-7.8

> ssh-keygen(1): write OpenSSH format private keys by default instead of using OpenSSL's PEM format

I think most supported distros have a newer version so up to you if you still wish to include it. I think it's ok to include as it's not too complicated but there shouldn't be too many people who need it.

tmds: I prefer to leave it out then. For things like this, I like users to create issues and upvote them to know they are important to include.

jborean93: It has been removed.

tmds:
Thank a lot @jborean93!

jborean93:
Thanks for the review and looking through the changes!

PR #211, opened 2024-08-13
Add logging support through Microsoft.Extensions.Logging.

tmds:


tmds:
@jborean93 fyi

The logging can also be helpful when contributing. Setting the log level to trace will print out the SSH packets being sent and received.

Issue #204, opened 2024-07-25
Ed25519 key support


jborean93:
Issue to track support for ED25519 private keys. I'm unsure what effort is involved to actually support this key type but as it's the recommended type used today for SSH supporting it in this library would be good.

jborean93:
Looks like https://github.com/dotnet/runtime/issues/63174 is a blocker for this unless a 3rd party dependency is taken up. Also mentioned by yourself in https://github.com/dotnet/runtime/issues/69130

tmds:
Yes, I agree we need Ed25519 and from the discussions on those issues, unfortunately, it seems .NET won't provide it. And even if it did, it wouldn't be supported on Windows.

jborean93:
I’m assuming the aim is to avoid any dependencies? That’s perfectly fine I just wanted to double check.

tmds:
Let's use Bouncy Castle (https://github.com/bcgit/bc-csharp) for implementing this.

jborean93:
I have no idea what the landscape is in .NET for cryptography libraries but [NSec](https://nsec.rocks) seemed to be another decent alternative.

tmds:
Bouncy Castle is well known. It is all managed code, and has not other dependencies.

For NSec, there is a dependency on a native library (libsodium) which I is something I'd like to avoid.

tmds:
This is part of 0.6.0 which was just uploaded to nuget.org.

PR #212, opened 2024-08-17
Add support for ed25519 private keys

jborean93:
Adds support for using ed25519 private key in user authentication. As .NET does not support ED25519 in the BCL it uses
BouncyCastle.Cryptography as a dependency for the key signing tasks.

Fixes https://github.com/tmds/Tmds.Ssh/issues/204


src/Tmds.Ssh/Ed25519PrivateKey.cs
```diff
@@ -0,0 +1,66 @@
...
+
+namespace Tmds.Ssh;
+
+sealed class Ed25519PrivateKey : PrivateKey
+{
+    private const int _privateKeySize = 32;
+
+    // Contains the private and public key as one block of bytes from the
+    // serialized OpenSSH key data.
+    private readonly byte[] _keyData;
```
jborean93, line 16: I used a single `byte[]` as that's how it is encoded in the raw OpenSSH key and the BouncyCastle API accepts an offset for the private and public key. Happy to just split that out in the parser and store it as 2 separate arrays if you prefer that.

tmds: Yes, please split into separate arrays for the private and public key.

src/Tmds.Ssh/Ed25519PrivateKey.cs
```diff
@@ -0,0 +1,66 @@
...
+
+using System;
+using System.Buffers;
+using Org.BouncyCastle.Math.EC.Rfc8032;
+
+namespace Tmds.Ssh;
+
+sealed class Ed25519PrivateKey : PrivateKey
+{
+    private const int _privateKeySize = 32;
```
tmds, line 12: nit: for a `const`, the convention is `PrivateKeySize`.

tmds:
Thanks for implementing this @jborean93!

src/Tmds.Ssh/Ed25519PrivateKey.cs
jborean93: Thanks I'll remember that for next time. Now that the array is already split in the parser it's no longer needed.

Issue #214, opened 2024-08-21
Cipher: chacha20-poly1305@openssh.com


tmds:
This is a modern cipher that we should support.

BCL does not include support for implementing this. We need a separate Poly and ChaCha20 handling.

cc @jborean93

tmds:
We currently only support AES GCM and this is implemented using the BCL `AesGcm` class. The underlying platform may not actually provide this algorithm ([Aes.Gcm.IsSupported](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.aesgcm.issupported?view=net-8.0) returns `false`).

For `chacha20-poly1305@openssh.com` we'll use a managed .NET implementation so it is always available.
This will enable the library to work on the platforms where AES GCM is not supported.

jborean93:
While implementing this would be ideal and now that BouncyCastle is a dependency it should be possible to achieve, AES GCM should be available on the three main platforms Windows, Linux, and macOS. MacOS seems to now have it available by default since .NET 8 https://learn.microsoft.com/en-us/dotnet/standard/security/cross-platform-cryptography#aes-gcm-and-chacha20poly1305-on-macos

> Starting in .NET 8, .NET on macOS will use the Apple's CryptoKit framework for AES-GCM and ChaCha20Poly1305.

tmds:
This is part of 0.6.0 which was just uploaded to nuget.org.

PR #215, opened 2024-08-24
Support chacha20-poly1305@openssh.com.

tmds:
Fixes https://github.com/tmds/Tmds.Ssh/issues/214.

@jborean93 fyi

### DIFF
--- a/src/Tmds.Ssh/SshClientSettings.Defaults.cs
+++ b/src/Tmds.Ssh/SshClientSettings.Defaults.cs
@@ -38,7 +38,7 @@ partial class SshClientSettings
     private readonly static List<Name> EmptyList = [];
     internal readonly static List<Name> SupportedKeyExchangeAlgorithms = [ AlgorithmNames.EcdhSha2Nistp256, AlgorithmNames.EcdhSha2Nistp384, AlgorithmNames.EcdhSha2Nistp521 ];
     internal readonly static List<Name> SupportedServerHostKeyAlgorithms = [ AlgorithmNames.EcdsaSha2Nistp521, AlgorithmNames.EcdsaSha2Nistp384, AlgorithmNames.EcdsaSha2Nistp256, AlgorithmNames.RsaSshSha2_512, AlgorithmNames.RsaSshSha2_256 ];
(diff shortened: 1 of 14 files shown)


tmds:
I currently put `chacha20-poly1305@openssh.com` first in `SupportedEncryptionAlgorithms` so all (`SshClientSettings`-based) tests are now running with this algorithm.

I will add new tests to this PR which test the ciphers separately.

src/Tmds.Ssh/SshClientSettings.Defaults.cs
```diff
@@ -63,6 +65,40 @@ private static IReadOnlyList<Credential> CreateDefaultCredentials()
         return credentials.AsReadOnly();
     }
 
+    private static List<Name> CreatePreferredEncryptionAlgorithms()
+    {
+        // Prefer AesGcm over ChaCha20Poly when the platform has AES instructions.
```
jborean93, line 70: Do we do this as AES with hardware acceleration is faster or is this what OpenSSH does as well?

src/Tmds.Ssh/SshClientSettings.Defaults.cs
```diff
@@ -63,6 +65,40 @@ private static IReadOnlyList<Credential> CreateDefaultCredentials()
...
+        bool addAesGcm = AesGcm.IsSupported;
+        bool hasAesInstructions =  System.Runtime.Intrinsics.X86.Aes.X64.IsSupported ||
```
jborean93, line 72: Shouldn't we be checking based on the process architecture and not all of them?

jborean93 (commented):
I can't really talk about the encryption code as I don't really understand chacha20poly1305 but the rest seems ok to me.

src/Tmds.Ssh/SshClientSettings.Defaults.cs
tmds: This is for ordering from fastest to slowest.

OpenSSH doesn't try to be smart. It has a fixed order.
The system config files enable the administrator to re-order the algorithms.

The `SshClient` can be created so it uses the OpenSSH config files (and follows their configuration).

src/Tmds.Ssh/SshClientSettings.Defaults.cs
tmds: I think we can just check all of them. When we're checking for a different architecture, the `IsSupported` will return `false`.

Issue #205, opened 2024-07-25
In memory key support


jborean93:
Right now the only way to use a key file is to specify the filepath of the key. It would be great to support a key that is either specified as a string or with the `PrivateKey` type itself. The latter would help to support things like using a key stored in a yubikey or other HSM like device.

tmds:
Sounds good.

API proposal:
```cs
PrivateKeyCredential(byte[] key);
```

jborean93:
As well as that I was potentially thinking more along the lines of

```csharp
PrivateKeyCredential(RSA key);
PrivateKeyCredential(ECDsa key);
PrivateKeyCredential(string content, bool isPath = true);
```

The first two would allow someone to specify the `Rsa/ECDsa` object directly allowing them to create it from their own source. You could simplify it and make the overload `AsymmetricAlgorithm` but then you'll still have to check the subtype is something we can utilise. Some benefits of offering this is so the key could be retrieved from;

+ The Windows certificate store attached to a cert
  + Windows can also store the private key by itself through a Win32 API and have it accessible by name but that's a bit more difficult
+ Build the key object through one of the many `Import*` methods on each type rather than be constrained by what this library can do
+ Source a key from Azure KeyVault or other HSM like devices like a Yubikey
+ Another way to load the key from memory depending on whatever the client desires

We could even show examples of some of the above which may require 3rd party APIs but at least they could implement it in their own code which calls Tmds.Ssh.

The last one I'm not sure on, it'll most be similar to your `byte[] key` option just a more convenient way to specify a PEM file by using a string. It may clutter the API a bit more, maybe it would be better as a static method like `static PrivateKeyCredential CreateFromString(string pem, string? password);` or something like that.

tmds:
```cs
PrivateKeyCredential(RSA key);
PrivateKeyCredential(ECDsa key);
```

I find these very advanced for an end-user to deal with, so I think we shouldn't add them unless there is a more clear demand.

```cs
PrivateKeyCredential(string content, bool isPath = true);
```

The `isPath` to change the meaning of `content` isn't one that is typically used in BCL APIs. We should have separate methods:

Let's start with:
```
PrivateKeyCredential(ReadOnlySpan<char> rawData, string? password = null)
```

Additional constructors or static create methods can be added later when users make requests for them.

tmds:
[...]

PR #223, opened 2024-09-05
Support using private keys that are not stored in files.

tmds:
This enables the key store use-case described in https://github.com/tmds/Tmds.Ssh/issues/205#issuecomment-2295601168.

This adds support for implementing such keys by deriving from `PrivateKeyCredential`.

We re-use the BCL `RSA` and `ECDsa` classes as abstractions for the key operations.

The implementation can be enhanced latter on by introducing our own abstractions for the key types. This would enable async key operations, and supporting key types the BCL has no abstraction for (like ED25519).

@jborean93 ptal.

### DIFF
--- a/src/Tmds.Ssh/PrivateKeyCredential.cs
+++ b/src/Tmds.Ssh/PrivateKeyCredential.cs
@@ -1,20 +1,94 @@
 // This file is part of Tmds.Ssh which is released under MIT.
 // See file LICENSE for full license details.
 
+using System.Security.Cryptography;
+
 namespace Tmds.Ssh;
 
-public sealed class PrivateKeyCredential : Credential
+public class PrivateKeyCredential : Credential
 {
-    internal string FilePath { get; }
+    internal string Identifier { get; }
 
-    internal Func<string?> PasswordPrompt { get; }
+    private Func<CancellationToken, ValueTask<Key>> LoadKey { get; }
 
-    public PrivateKeyCredential(string path, string? password = null) : this(path, () => password)
+    public PrivateKeyCredential(string path, string? password = null, string? identifier = null) :
+        this(path, () => password, identifier)
     { }
 
-    public PrivateKeyCredential(string path, Func<string?> passwordPrompt)
+    public PrivateKeyCredential(string path, Func<string?> passwordPrompt, string? identifier = null) :
+        this(LoadKeyFromFile(path ?? throw new ArgumentNullException(nameof(path)), passwordPrompt), identifier ?? path)
+    { }
+
+    // Allows the user to implement derived classes that represent a private key.
+    protected PrivateKeyCredential(Func<CancellationToken, ValueTask<Key>> loadKey, string identifier)
     {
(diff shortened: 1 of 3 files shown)


src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -1,20 +1,94 @@
...
+
 namespace Tmds.Ssh;
 
-public sealed class PrivateKeyCredential : Credential
+public class PrivateKeyCredential : Credential
 {
-    internal string FilePath { get; }
+    internal string Identifier { get; }
+
+    private Func<CancellationToken, ValueTask<Key>> LoadKey { get; }
```
jborean93, line 12: `Key` seems to be `nullable` based on https://github.com/tmds/Tmds.Ssh/pull/223/files#diff-e8a24fa45fee86b727c97d8b80fe73f7113dcd748f6a7f871031b1e1d0d5f407R25-R30

src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -1,20 +1,94 @@
...
+        internal PrivateKey? PrivateKey { get; }
+
+        public Key(RSA rsa)
+        {
+            PrivateKey = new RsaPrivateKey(rsa);
+        }
+
+        public Key(ECDsa ecdsa)
+        {
+            ECParameters parameters = ecdsa.ExportParameters(includePrivateParameters: false);
```
jborean93, line 57: Is this logic duplicated elsewhere when we do the key loading as a file? Maybe this should be a static method on `ECDsaPrivateKey` to consolidate the logic.

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: This `Key` is a `struct`, the nullable one is its nullable `PrivateKey` property that gets returned by the `LoadKeyAsync` (when `LoadKey` returns `default(Key)`).

src/Tmds.Ssh/PrivateKeyCredential.cs
tmds: I don't think we have something similar to this that we might put in a method.

tmds: I think you are referring to `TryParseOpenSshEcdsaKey`. I'll take a closer look and see if it makes sense to share some code.

tmds: I took a look and it's hard to re-use something here because these methods are dealing with different inputs. `TryParseOpenSshEcdsaKey` starts with a key identifier, and a curve name, whereas here we start with an `ECDsa`.

tmds:
@jborean93 would you be interested in writing up a code snippit that we can add to a `howto.md` that demonstrates how this may be used to implement an `AzureKeyCredential`?

jborean93:
Sure when I get time I'll look at adding that!

PR #226, opened 2024-09-15
Add Azure Key example and fix ECDSA OID check

jborean93:
Adds an example of using this library with a private key hosted in Azure Key Vault. This also includes a fix for loading an ephemeral ECDSA key from the .NET object by properly checking the curve OID value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,14 @@ $ dotnet run
 hello world!
 ```
 
+## Examples
+
+The following are some example projects that show how Tmds.Ssh can be used:
+
+- [scp](./examples/scp) - SCP client to copy/fetch files
+- [ssh](./examples/ssh) - SSH client
+- [azure_key](./examples/azure_key) - SSH client with private keys stored in Azure Key Vault.
+
 ## API
 
 ```cs
@@ -500,14 +508,6 @@ Under these levels, the logged messages may include:
 - file paths (including those of private keys)
 
 The `Debug` and `Trace` loglevels should not be used in production. Under the `Trace` level all packets are logged. This will expose sensitive data related to the SSH connection and the application itself.
-
-## Examples
-
-The following are some example projects that show how Tmds.Ssh can be used:
-
-- [scp](./examples/scp) - SCP client to copy/fetch files
-- [ssh](./examples/ssh) - SSH client
-- [azure_key](./examples/azure_key) - SSH client with private keys stored in Azure Key Vault.
 
 ## CI Feed
 
--- a/README.md
+++ b/README.md
@@ -501,6 +501,14 @@ Under these levels, the logged messages may include:
 
 The `Debug` and `Trace` loglevels should not be used in production. Under the `Trace` level all packets are logged. This will expose sensitive data related to the SSH connection and the application itself.
 
+## Examples
+
+The following are some example projects that show how Tmds.Ssh can be used:
+
+- [scp](./examples/scp) - SCP client to copy/fetch files
+- [ssh](./examples/ssh) - SSH client
+- [azure_key](./examples/azure_key) - SSH client with private keys stored in Azure Key Vault.
+
 ## CI Feed
 
 You can obtain packages from the CI NuGet feed: https://www.myget.org/F/tmds/api/v3/index.json.
--- a/examples/azure_key/AzureECDsaKey.cs
+++ b/examples/azure_key/AzureECDsaKey.cs
@@ -1,0 +1,37 @@
+using Azure.Security.KeyVault.Keys.Cryptography;
+using System;
+using System.Security.Cryptography;
+using AzureSignatureAlgorithm = Azure.Security.KeyVault.Keys.Cryptography.SignatureAlgorithm;
+
+namespace Tmds.Ssh.AzureKeyExample;
+
+sealed class AzureECDsaKey : ECDsa
+{
+    private readonly CryptographyClient _cryptoClient;
+    private readonly ECParameters _publicParameters;
+    private readonly AzureSignatureAlgorithm _signatureAlgorithm;
+
+    public AzureECDsaKey(CryptographyClient client, ECParameters publicParameters, AzureSignatureAlgorithm signatureAlgorithm)
+    {
+        KeySizeValue = publicParameters.Q.X!.Length * 8;
+        _cryptoClient = client;
+        _publicParameters = publicParameters;
+        _signatureAlgorithm = signatureAlgorithm;
+    }
+
+    public override ECParameters ExportParameters(bool includePrivateParameters)
+    {
+        if (includePrivateParameters)
+        {
+            throw new CryptographicException("Cannot export private parameters");
+        }
+
+        return _publicParameters;
+    }
+
+    public override byte[] SignHash(byte[] hash)
+        => _cryptoClient.Sign(_signatureAlgorithm, hash).Signature;
+
+    public override bool VerifyHash(byte[] hash, byte[] signature)
+        => throw new NotImplementedException();
+}
--- a/examples/azure_key/AzureECDsaKey.cs
+++ b/examples/azure_key/AzureECDsaKey.cs
@@ -1,6 +1,7 @@
 using Azure.Security.KeyVault.Keys.Cryptography;
 using System;
 using System.Security.Cryptography;
+using System.Threading;
 using AzureSignatureAlgorithm = Azure.Security.KeyVault.Keys.Cryptography.SignatureAlgorithm;
 
 namespace Tmds.Ssh.AzureKeyExample;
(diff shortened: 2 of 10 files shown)


jborean93:
@tmds let me know if you want to split out the OID bugfix into a separate PR so your review on the Azure bits don't hold up your next release.

examples/azure_key/AzureECDsaKey.cs
```diff
@@ -0,0 +1,37 @@
...
+        if (includePrivateParameters)
+        {
+            throw new CryptographicException("Cannot export private parameters");
+        }
+
+        return _publicParameters;
+    }
+
+    public override byte[] SignHash(byte[] hash)
+        => _cryptoClient.Sign(_signatureAlgorithm, hash).Signature;
```
tmds, line 33: Can you change this to `_cryptoClient.SignAsync(_signatureAlgorithm, hash, _cancellationToken).GetAwaiter().GetResult().Signature;`, where `_cancellationToken` is the one we get from the `LoadKey` function and gets passed to this class through the constructor?

examples/azure_key/AzureKeyCredential.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Tmds.Ssh.AzureKeyExample;
+
+sealed class AzureKeyCredential : PrivateKeyCredential
+{
+    public AzureKeyCredential(string vaultName, string keyName) :
+        base((c) => GetAzureKey(vaultName, keyName, c), $"Azure Key Vault: {vaultName}/{keyName}")
+    { }
+
+    public static async ValueTask<string> GetAzurePubKey(string vaultName, string keyName)
```
tmds, line 21: I think this method and the related `GetRsaPubKey`/`GetEcdsaPubKey` are for working with the public side of the key and they are unrelated to the private key handling needed for the library. Can you move them into `Program.cs` or into a separate class.

examples/azure_key/AzureKeyCredential.cs
```diff
@@ -0,0 +1,137 @@
...
+                .ExportParameters(false);
+            return GetEcdsaPubKey(pubParams);
+        }
+        else
+        {
+            throw new NotImplementedException($"Unsupported Azure key type {key.KeyType}");
+        }
+    }
+
+    private static async ValueTask<Key> GetAzureKey(string vaultName, string keyName, CancellationToken cancellationToken = default)
```
tmds, line 43: Can you use the `Async` suffix for anything that returns a `Task`/`ValueTask`?

examples/azure_key/AzureKeyCredential.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        (TokenCredential cred, KeyVaultKey key) = await GetKeyVaultKey(vaultName, keyName);
+
+        if (key.KeyType == KeyType.Rsa)
+        {
+            CryptographyClient azureClient = new CryptographyClient(key.Id, cred);
+            RSAParameters pubParams = key.Key.ToRSA(includePrivateParameters: false)
+                .ExportParameters(false);
+
+            return new Key(new AzureRsaKey(azureClient, pubParams));
```
tmds, line 53: side-note: it was intentional that the derived class would be able to return keys of different types as you are doing here.

examples/azure_key/AzureKeyCredential.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using AzureSignatureAlgorithm = Azure.Security.KeyVault.Keys.Cryptography.SignatureAlgorithm;
+
+namespace Tmds.Ssh.AzureKeyExample;
+
+sealed class AzureKeyCredential : PrivateKeyCredential
+{
+    public AzureKeyCredential(string vaultName, string keyName) :
+        base((c) => GetAzureKey(vaultName, keyName, c), $"Azure Key Vault: {vaultName}/{keyName}")
```
tmds, line 18: nit/suggestion, may be we use a uri/urn type of identifier. `azure:{vaultName}:{keyName}`

src/Tmds.Ssh/PrivateKeyCredential.cs
```diff
@@ -119,6 +119,9 @@ internal Key(PrivateKey key)
         {
             PrivateKey = key;
         }
+
+        private static bool OidEquals(Oid oidA, Oid oidB)
+            => oidA.Value is not null && oidB.Value is not null && oidA.Value == oidB.Value;
```
tmds, line 124: Thank you for fixing this!

test/Tmds.Ssh.Tests/PrivateKeyCredentialTests.cs
```diff
@@ -155,6 +163,52 @@ await RunWithKeyConversion(_sshServer.TestUserIdentityFileEcdsa521, async (strin
...
+
+    [Fact]
+    public async Task FailEcdsaKeyWithInvalidCurve()
+    {
+        const string expected = "Curve 'brainpoolP256r1' is not known.";
+
+        var credential = new ECDsaKeyCredential(ECCurve.NamedCurves.brainpoolP256r1);
+        NotSupportedException exc = await Assert.ThrowsAsync<NotSupportedException>(async () => await credential.LoadKeyAsync(default));
+        Assert.Equal(expected, exc.Message);
+    }
```
tmds, line 210: Good tests!

side-note: I'd like us to update the existing tests in this class to speed it up by: pre-generating keys and storing them as const strings (like the existing `PrivateRsaKey`) and trying to load them and check values (instead of doing a full connect).

tmds:
Great example, @jborean93! I've added some minor suggestions.

README.md
```diff
@@ -501,6 +501,14 @@ Under these levels, the logged messages may include:
 
 The `Debug` and `Trace` loglevels should not be used in production. Under the `Trace` level all packets are logged. This will expose sensitive data related to the SSH connection and the application itself.
 
+## Examples
```
tmds, line 504: You can move this up so it comes right after _Getting Started_.

examples/azure_key/AzureECDsaKey.cs
jborean93: I can, is it a guarantee that the cancellation token is going to be the same going forward?

examples/azure_key/AzureKeyCredential.cs
jborean93: Sure thing

examples/azure_key/AzureKeyCredential.cs
jborean93: Sure thing.

src/Tmds.Ssh/PrivateKeyCredential.cs
jborean93: Yw

test/Tmds.Ssh.Tests/PrivateKeyCredentialTests.cs
jborean93: I can certainly update these new tests to load the key from a serialized string. This particular test doesn't do a full connection and while the others do so I think there is some merit for actually testing the end to end connection. Ultimately it's up to you whether you prefer the shorter and quicker test setup which certainly has its merits.

README.md
jborean93: Will do thanks

examples/azure_key/AzureECDsaKey.cs
tmds: The cancellationtoken is for the scope of the connect, and the key that gets created here has a shorted lifetime than the connect, so it's appropriate to store it in the constructor argument.

examples/azure_key/AzureKeyCredential.cs
tmds: I'd like the sample to focus on how to use `PrivateKeyCredential` with `LoadKey `and move code not needed for that out of the derived class.

If there is some code you want to share, you could make the `PrivateKeyCredential` `internal`.

examples/azure_key/AzureKeyCredential.cs
tmds: No change required. This is nicely demonstrating how different key types can be loaded by a derived class!

test/Tmds.Ssh.Tests/PrivateKeyCredentialTests.cs
tmds: The tests added here are good.

> Ultimately it's up to you whether you prefer the shorter and quicker test setup which certainly has its merits.

Yes, I'd like to put some effort into making the tests run faster (without sacrificing coverage).

examples/azure_key/AzureECDsaKey.cs
jborean93: Sounds good to me

examples/azure_key/AzureKeyCredential.cs
jborean93: Makes sense, I've moved the code to `Program.cs`. I didn't want to remove it entirely because there is no other way to extract the public key for the `authorized_keys` making it harder to use.

examples/azure_key/AzureKeyCredential.cs
jborean93: Fantastic, just wanted to confirm!

tmds:
Thanks @jborean93!

jborean93:
Thanks for the review

PR #252, opened 2024-11-24
Remove Internal Constructor of ExecuteOptions

EX-EXE:
The ExecuteOptions constructor was marked as internal, which prevented it from being specified as an argument for ExecuteAsync in external applications.

To resolve this issue, I have removed the internal constructor of ExecuteOptions.


tmds:
Thanks!

tmds:
This is part of 0.9.0 which was just uploaded to nuget.org.

PR #264, opened 2024-12-08
Add support for PKCS#8 private key

scott-xu:


tmds:
hi @scott-xu, while keys can be in the PKCS#8 format, I think this is uncommon for SSH keys. As described in the README, the library focuses on OpenSSH formats. Other formats can be added but a "strong" case has to be made for it first.

I hope this makes sense. I'm going to close the PR. Please open an issue before making a PR (unless you are fixing an obvious bug).

PR #266, opened 2024-12-09
Support downloading to a Stream (#261)

### LINKED ISSUE #261: Support downloading to a Stream
please add the possibility to download file to a Stream
SSH.Net has this possibility, but is way to slow
i adapted your library locally adding this feature to test and is way faster
but it would be better to depend on a nuget package instead of managing it myself
let me know if you want me to make pull request with my adaptions for this use case

yokrysty:
closes #261

### DIFF
--- a/src/Tmds.Ssh/SftpChannel.cs
+++ b/src/Tmds.Ssh/SftpChannel.cs
@@ -1052,7 +1052,13 @@ sealed partial class SftpChannel : IDisposable
     public ValueTask DownloadFileAsync(string remoteFilePath, string localFilePath, bool overwrite = false, CancellationToken cancellationToken = default)
         => DownloadFileAsync(remoteFilePath, localFilePath, length: null, overwrite, permissions: null, throwIfNotFound: true, cancellationToken);
 
-    private async ValueTask DownloadFileAsync(string remotePath, string localPath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+    public ValueTask DownloadFileAsync(string remotePath, string localFilePath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+        => DownloadFileAsync(remotePath, (object)localFilePath, length, overwrite, permissions, throwIfNotFound, cancellationToken);
+
+    public ValueTask DownloadFileAsync(string remotePath, Stream destination, long? length, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+        => DownloadFileAsync(remotePath, destination, length, false, permissions, throwIfNotFound, cancellationToken);
+
+    private async ValueTask DownloadFileAsync(string remotePath, object destination, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
     {
         // Call GetAttributesAsync in parallel with OpenFileAsync.
         // We don't need to pass the CancellationToken since GetAttributesAsync will complete before the awaited OpenFileAsync completes.
@@ -1079,27 +1085,49 @@ sealed partial class SftpChannel : IDisposable
             permissions = attributes.Permissions;
         }
 
-        using FileStream localFile = OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions!.Value);
+        Stream? dstStream = null;
+        if (destination is string localPath)
+        {
+            dstStream = OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions.Value);
+        }
+        else if (destination is Stream stream)
+        {
+            dstStream = stream;
+        }
+
+        Debug.Assert(dstStream is not null);
 
         ValueTask previous = default;
-
         CancellationTokenSource? breakLoop = length > 0 ? new() : null;
 
-        int maxPayload = MaxReadPayload;
-        for (long offset = 0; offset < length; offset += maxPayload)
+        Debug.Assert(breakLoop is not null);
+
+        try
         {
-            Debug.Assert(breakLoop is not null);
-            if (!breakLoop.IsCancellationRequested)
+            int maxPayload = MaxReadPayload;
+            for (long offset = 0; offset < length; offset += maxPayload)
             {
-                await s_downloadBufferSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
-                long remaining = length.Value - offset;
-                previous = CopyBuffer(previous, offset, remaining > maxPayload ? maxPayload : (int)remaining);
+                if (!breakLoop.IsCancellationRequested)
+                {
+                    await s_downloadBufferSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+                    long remaining = length.Value - offset;
+                    previous = CopyBuffer(previous, offset, remaining > maxPayload ? maxPayload : (int)remaining);
+                }
+            }
+
+            await previous.ConfigureAwait(false);
+
+            await remoteFile.CloseAsync(cancellationToken).ConfigureAwait(false);
+        }
+        finally
+        {
+            if (destination is string)
+            {
+                dstStream.Dispose();
             }
         }
 
-        await previous.ConfigureAwait(false);
-
-        await remoteFile.CloseAsync(cancellationToken).ConfigureAwait(false);
+        return;
 
         async ValueTask CopyBuffer(ValueTask previousCopy, long fileOffset, int length)
         {
@@ -1135,7 +1163,7 @@ sealed partial class SftpChannel : IDisposable
                     await previousCopy.ConfigureAwait(false);
                 }
 
-                localFile.Write(buffer.AsSpan(0, length));
+                dstStream.Write(buffer.AsSpan(0, length));
             }
             catch
             {
--- a/src/Tmds.Ssh/SftpChannel.cs
+++ b/src/Tmds.Ssh/SftpChannel.cs
@@ -949,7 +949,7 @@ sealed partial class SftpChannel : IDisposable
                         break;
                     case UnixFileType.RegularFile:
                         lastDirectory = EnsureParentDirectory(lastDirectory, item.LocalPath);
-                        onGoing.Enqueue(DownloadFileAsync(item.RemotePath, item.LocalPath, item.Length, overwrite, item.Permissions, throwIfNotFound: false, cancellationToken));
+                        onGoing.Enqueue(DownloadFileAsync(item.RemotePath, item.LocalPath, null, item.Length, overwrite, item.Permissions, throwIfNotFound: false, cancellationToken));
                         break;
                     case UnixFileType.SymbolicLink:
                         lastDirectory = EnsureParentDirectory(lastDirectory, item.LocalPath);
@@ -1055,9 +1055,6 @@ sealed partial class SftpChannel : IDisposable
     public ValueTask DownloadFileAsync(string remotePath, Stream destination, CancellationToken cancellationToken = default)
         => DownloadFileAsync(remotePath, localPath: null, destination, length: null, overwrite: false, permissions: null, throwIfNotFound: true, cancellationToken);
 
-    private ValueTask DownloadFileAsync(string remotePath, string? localPath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
-        => DownloadFileAsync(remotePath, localPath, null, length, overwrite, permissions, throwIfNotFound, cancellationToken);
-
     private async ValueTask DownloadFileAsync(string remotePath, string? localPath, Stream? destination, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
     {
         // Call GetAttributesAsync in parallel with OpenFileAsync.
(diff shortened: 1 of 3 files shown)


src/Tmds.Ssh/SftpClient.cs
```diff
@@ -378,6 +378,12 @@ public async ValueTask DownloadFileAsync(string remoteFilePath, string localFile
         await channel.DownloadFileAsync(remoteFilePath, localFilePath, overwrite, cancellationToken);
     }
 
+    public async ValueTask DownloadFileAsync(string remoteFilePath, Stream destination, CancellationToken cancellationToken = default)
+    {
+        var channel = await GetChannelAsync(cancellationToken).ConfigureAwait(false);
+        await channel.DownloadFileAsync(remoteFilePath, destination, null, null, true, cancellationToken);
```
tmds, line 384: Please remove the constant arguments (`null, null, true`) from the SftpChannel method being called here.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1052,7 +1052,13 @@ private async ValueTask DownloadLinkAsync(string remotePath, string localPath, b
...
 
-    private async ValueTask DownloadFileAsync(string remotePath, string localPath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+    public ValueTask DownloadFileAsync(string remotePath, string localFilePath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+        => DownloadFileAsync(remotePath, (object)localFilePath, length, overwrite, permissions, throwIfNotFound, cancellationToken);
+
+    public ValueTask DownloadFileAsync(string remotePath, Stream destination, long? length, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+        => DownloadFileAsync(remotePath, destination, length, false, permissions, throwIfNotFound, cancellationToken);
+
+    private async ValueTask DownloadFileAsync(string remotePath, object destination, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
```
tmds, line 1061: Replace the `object destination` arg by separate `string? localFilePath, Stream? destination` args.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1079,27 +1085,49 @@ private async ValueTask DownloadFileAsync(string remotePath, string localPath, l
             permissions = attributes.Permissions;
         }
 
-        using FileStream localFile = OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions!.Value);
+        Stream? dstStream = null;
```
tmds, line 1088: We've lost the `using`.

You can do something like:
```cs
using FileStream? localFile = localPath is null ? null : OpenFileStream(...);
stream ??= localFile;
```

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1079,27 +1085,49 @@ private async ValueTask DownloadFileAsync(string remotePath, string localPath, l
...
 
-        ValueTask previous = default;
+        Debug.Assert(dstStream is not null);
 
+        ValueTask previous = default;
         CancellationTokenSource? breakLoop = length > 0 ? new() : null;
 
-        int maxPayload = MaxReadPayload;
-        for (long offset = 0; offset < length; offset += maxPayload)
+        Debug.Assert(breakLoop is not null);
```
tmds, line 1103: Can you move this back? `breakLoop` can be null here (when `length` is zero).

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1135,7 +1163,7 @@ async ValueTask CopyBuffer(ValueTask previousCopy, long fileOffset, int length)
                     await previousCopy.ConfigureAwait(false);
                 }
 
-                localFile.Write(buffer.AsSpan(0, length));
+                dstStream.Write(buffer.AsSpan(0, length));
```
tmds, line 1166: Higher up, when `stream` got assigned, can you add:
```cs
bool writeSync = stream is FileStream or MemoryStream;
```
and then here:
```cs
if (writeSync)
{
    stream.Write(buffer.AsSpan(0, length));
}
else
{
    await stream.WriteAsync(buffer.AsMemory(0, length), cancellationToken);
}
```

test/Tmds.Ssh.Tests/SftpClientTests.cs
```diff
@@ -686,13 +686,19 @@ public async Task UploadDownloadFile(int fileSize)
             // Download
             await sftpClient.DownloadFileAsync(remotePath, destinationPath);
 
+            var downloadStream = new MemoryStream();
```
tmds, line 689: Instead of adding it here, can you add a separate test. Something like:

```cs
[Fact]
DownloadFileToStream()
{
  using var sftpClient = await _sshServer.CreateSftpClientAsync(sftpExtensions);

  (string sourceFileName, byte[] sourceData) = await CreateRemoteFileWithRandomDataAsync(sftpClient, fileSize: 100);

  ...
}
```

src/Tmds.Ssh/SftpChannel.cs
yokrysty: i thought about exactly that but i think the fileOffset is need to the destination or else everything will be messed up, am i wrong?

test/Tmds.Ssh.Tests/SftpClientTests.cs
yokrysty: will do

src/Tmds.Ssh/SftpChannel.cs
tmds: Yes, I noticed the `finally` later as I read through the PR.
Can you change to adopt the pattern suggested above?

src/Tmds.Ssh/SftpChannel.cs
yokrysty: will do, so you like 2 params instead of 1 :)

src/Tmds.Ssh/SftpChannel.cs
yokrysty: yes, but i really don't understand the point to be in the loop since everything is assigned before the loop and nothing is reassigned in the loop

src/Tmds.Ssh/SftpChannel.cs
yokrysty: yes will do

src/Tmds.Ssh/SftpChannel.cs
tmds: It works well with https://github.com/tmds/Tmds.Ssh/pull/266/files/72c23ea76d683e464dee46cd08b191af896f57f4#diff-e75f1a3245a183befaa84212ba7019b20d19d6ea2425d893f2e5cf8d5ec84f5f because we'll end up using the `Stream` argument in both cases.

src/Tmds.Ssh/SftpChannel.cs
yokrysty: ah yes, my bad

src/Tmds.Ssh/SftpChannel.cs
yokrysty: ah yes i was thinking to remove that when async that's why i ended up needing fileOffset

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1079,10 +1085,11 @@ private async ValueTask DownloadFileAsync(string remotePath, string localPath, l
             permissions = attributes.Permissions;
         }
 
-        using FileStream localFile = OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions!.Value);
+        using FileStream? localFile = localPath is null ? null : OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions.Value);
+        destination ??= localFile;
+        var writeSync = destination is FileStream or MemoryStream;
```
tmds, line 1090: Please change `var` to `bool` here.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1101,6 +1108,8 @@ private async ValueTask DownloadFileAsync(string remotePath, string localPath, l
 
         await remoteFile.CloseAsync(cancellationToken).ConfigureAwait(false);
 
+        return;
+
```
tmds, line 1112: Can you remove these two lines?

test/Tmds.Ssh.Tests/SftpClientTests.cs
```diff
@@ -708,6 +708,18 @@ public async Task UploadDownloadFile(int fileSize)
...
+    [Fact]
+    public async Task DownloadFileToStream()
+    {
+        using var sftpClient = await _sshServer.CreateSftpClientAsync();
+        var (sourceFileName, sourceData) = await CreateRemoteFileWithRandomDataAsync(sftpClient, length: 100);
+
+        await using var downloadStream = new MemoryStream();
+        await sftpClient.DownloadFileAsync(sourceFileName, downloadStream);
+
+        Assert.True(downloadStream.ToArray().SequenceEqual(sourceData));
```
tmds, line 720: This should work: `Assert.Equal(sourceData, downloadStream.ToArray());`.

tmds:
@yokrysty thanks for the updates! I have some more minor request and then this will be good to merge.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1050,9 +1050,15 @@ private async ValueTask DownloadLinkAsync(string remotePath, string localPath, b
...
 
     public ValueTask DownloadFileAsync(string remoteFilePath, string localFilePath, bool overwrite = false, CancellationToken cancellationToken = default)
-        => DownloadFileAsync(remoteFilePath, localFilePath, length: null, overwrite, permissions: null, throwIfNotFound: true, cancellationToken);
+        => DownloadFileAsync(remoteFilePath, localFilePath, destination: null, length: null, overwrite, permissions: null, throwIfNotFound: true, cancellationToken);
 
-    private async ValueTask DownloadFileAsync(string remotePath, string localPath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
+    public ValueTask DownloadFileAsync(string remotePath, Stream destination, CancellationToken cancellationToken = default)
+        => DownloadFileAsync(remotePath, localPath: null, destination, length: null, overwrite: false, permissions: null, throwIfNotFound: true, cancellationToken);
+
+    private ValueTask DownloadFileAsync(string remotePath, string? localPath, long? length, bool overwrite, UnixFilePermissions? permissions, bool throwIfNotFound, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
yokrysty: i didn't wanted to interfere with the existing code, but ok if i have your approval

src/Tmds.Ssh/SftpChannel.cs
yokrysty: i am a fan of var, but i will change it to your style

src/Tmds.Ssh/SftpChannel.cs
yokrysty: that was a ReSharper recommendation when using functions in functions, but i will remove it

yokrysty:
@tmds great, almost there, after this i have prepared an implementation for uploading from a stream
i will create a new issue with a PR and i hope you will like the changes

tmds:
> After this i have prepared an implementation for uploading from a stream i will create a new issue with a PR

Note that this one is a bigger challenge. In contrast to the `Stream` being downloaded to here, the source for an upload isn't read sequentially to enable pipelining the writes. We want to maintain the pipelining for `FileStream` and preferably also have it for `MemoryStream`. For an async `Stream` case, pipelining is not possible. Please create an issue for the feature, and if you make a PR for it, try to handle these different cases.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -949,7 +949,7 @@ public async ValueTask DownloadDirectoryEntriesAsync(string remoteDirPath, strin
                         break;
                     case UnixFileType.RegularFile:
                         lastDirectory = EnsureParentDirectory(lastDirectory, item.LocalPath);
-                        onGoing.Enqueue(DownloadFileAsync(item.RemotePath, item.LocalPath, item.Length, overwrite, item.Permissions, throwIfNotFound: false, cancellationToken));
+                        onGoing.Enqueue(DownloadFileAsync(item.RemotePath, item.LocalPath, null, item.Length, overwrite, item.Permissions, throwIfNotFound: false, cancellationToken));
```
tmds, line 952: Please include the arg name: `destination: null`.

src/Tmds.Ssh/SftpChannel.cs
```diff
@@ -1079,10 +1082,11 @@ private async ValueTask DownloadFileAsync(string remotePath, string localPath, l
             permissions = attributes.Permissions;
         }
 
-        using FileStream localFile = OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions!.Value);
+        using FileStream? localFile = localPath is null ? null : OpenFileStream(localPath, overwrite ? FileMode.Create : FileMode.CreateNew, FileAccess.Write, FileShare.None, permissions.Value);
+        destination ??= localFile;
```
tmds, line 1086: Add a `Debug.Assert(destination is not null)` to eliminate the nullability warnings.

src/Tmds.Ssh/SftpChannel.cs
yokrysty: done

src/Tmds.Ssh/SftpChannel.cs
yokrysty: done

tmds:
Thanks @yokrysty!

tmds:
If you'd like to use this feature before it's released, you can use a package from the CI feed: https://github.com/tmds/Tmds.Ssh?tab=readme-ov-file#ci-feed.

tmds:
This is part of 0.9.0 which was just uploaded to nuget.org.

Issue #267, opened 2024-12-10
Support uploading from a Stream


yokrysty:
title is self explanatory, the counterpart of #261

tmds:
API:
```cs
class SftpClient
{
  ValueTask UploadFileAsync(Stream source, string remoteFilePath, CancellationToken cancellationToken);
  ValueTask UploadFileAsync(Stream source, string remoteFilePath, bool overwrite = false, UnixFilePermissions? createPermissions = null, CancellationToken cancellationToken = default);
```

yokrysty:
you can check what i did before the doing a PR, since i am not sure it's the way you wanted

tmds:
hi @yokrysty, I took a look and I'd like to do it a bit differently. I'm going to tackle this one myself and I'll cc you on the PR so you can take a look and give feedback if you want.

tmds:
This is part of 0.9.0 which was just uploaded to nuget.org.

PR #271, opened 2024-12-11
SftpClient: support uploading file from Stream.

tmds:
Fixes https://github.com/tmds/Tmds.Ssh/issues/267.

cc @yokrysty

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ class SftpClient : IDisposable
 
   ValueTask UploadFileAsync(string localFilePath, string remoteFilePath, CancellationToken cancellationToken);
   ValueTask UploadFileAsync(string localFilePath, string remoteFilePath, bool overwrite = false, UnixFilePermissions? createPermissions, CancellationToken cancellationToken = default);
+  ValueTask UploadFileAsync(Stream source, string remoteFilePath, CancellationToken cancellationToken);
+  ValueTask UploadFileAsync(Stream source, string remoteFilePath, bool overwrite = false, UnixFilePermissions? createPermissions = null, CancellationToken cancellationToken = default);
   ValueTask UploadDirectoryEntriesAsync(string localDirPath, string remoteDirPath, CancellationToken cancellationToken = default);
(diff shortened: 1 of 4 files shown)


yokrysty:
oh yes i see, seems better you tackled the async writes which i didn't
for my use cases was enough

some questions:
i still don't understand the async usage on streams because you have await previousCopy.ConfigureAwait(false), so everything goes one after another (i also addressed this one in my pull request, but probably you didn't saw it)
i am thinking more like writing chunks at different offsets in parallel

in the method GetPermissionsForFile the Default has an additional & ~PretendUMask
which is skipped when having a Stream and i don't know what is doing, is it needed or not?
in my changes the reason i used the GetPermissionsForFile is because you had there implementation for Windows (which is currently not finished) and Linux

tmds:
> i still don't understand the async usage on streams because you have await previousCopy.ConfigureAwait(false), so everything goes one after another (i also addressed this one in my pull request, but probably you didn't saw it)
i am thinking more like writing chunks at different offsets in parallel

 Here the `previousCopy.ConfigureAwait` happens at the end, so the writes are happening in parallel (when `pipelineSyncWrites`).

In the `!pipelineSyncWrites` case there is some room for optimization.

> in the method GetPermissionsForFile the Default has an additional & ~PretendUMask
which is skipped when having a Stream and i don't know what is doing, is it needed or not?

By lack of information `GetPermissionsForFile` assumes files on Windows aren't writable for 'other'.
Note that when the file/directory is created on the server, it will also apply a [umask](https://man7.org/linux/man-pages/man1/umask.1p.html) that filters out permission. Normally, it will filter out the 'other write' permission (and often also the 'group write' permission).

yokrysty:
still in don't understand how it's not everything messed up at the destination without knowing the offset where to write
but i think because you await CopyToAsync for upload and await WriteAsync for download is doing everything in order

tmds:
> how it's not everything messed up at the destination

In the `!pipelineSyncWrites` case, CopyToAsync does serial writing.
In the `pipelineSyncWrites` case, each `CopyBuffer` works on a specific `offset`. The SFTP SSH write operation will write specifically at that offset.

PR #272, opened 2024-12-12
[Breaking] Drop support for RSA PKCS#1 PEM keys in favor of the OpenSSH format.

tmds:
As described in the README, this library embraces OpenSSH format when possible.

This drops support for the RSA PKCS#1 PEM ('RSA PRIVATE KEY') format which was supported because the library author was using this format for a key.

If you are using a key in this format, it can be converted by running:

   ssh-keygen -p -f <key_file>

The OpenSSH format has been supported in OpenSSH since 2014 and became the default format used by 'ssh-keygen' in 2018.


tmds:
@jborean93 fyi.

yokrysty:
well this change broke everything for me
i used the command to convert the key and now i get this error:
_The specified RSA parameters are not valid. Exponent and Modulus are required. If D is present, it must have the same length as Modulus. If D is present, P, Q, DP, DQ, and InverseQ are required and must have half the length of Modulus, rounded up, otherwise they must be omitted._
i also tried to use puttygen -> Conversions -> Export OpenSSH Key (force new file format) with the same result
tried more tools on windows and linux without luck
i did some debugging and ended up finding that the InverseQ is not the same after the conversion
image
now since the old format is not supported anymore i cannot use the library

so you have an ideea what is going on?

tmds:
Sorry for breaking your app @yokrysty.

I assume the converted key is working well with other tools, like `ssh`.

From the debug info you provided, I think we're not meeting the length requirements that the .NET class is expecting:

> If D is present, it must have the same length as Modulus. If D is present, P, Q, DP, DQ, and InverseQ are required and must have half the length of Modulus, rounded up, otherwise they must be omitted.

This code needs to be updated for that:

https://github.com/tmds/Tmds.Ssh/blob/f4501c56b65a916c09676e94a03830a6d6bcbd9d/src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs#L171-L191

With your key only the InverseQ is not the expected value. Can you try updating the above code by setting the `minLength` on `ReadMPIntAsByteArray` when we read `inverseQ` so we meet the length requirement, and see if that makes things work?

yokrysty:
yes it works with:
`byte[] inverseQ = reader.ReadMPIntAsByteArray(isUnsigned: true, minLength: modulus.Length / 2);`

tmds:
Great! I'll look at fixing this and include it in a patch release in the coming day(s).

yokrysty:
forgot to mention that the app is running on Windows .net inside IIS
i see that on Windows RSA.Create() is using new RSABCrypt() which seems a wrapper around some Windows native dll

tmds:
@yokrysty the fix is in 0.9.1 which I've just uploaded to nuget.org.

PR #273, opened 2024-12-12
PrivateKeyCredential: fix issue when loading some RSA keys.

tmds:
We were not meeting some length expectations for parameters provided to the .NET RSA class.

Fixes https://github.com/tmds/Tmds.Ssh/pull/272#issuecomment-2539716021.

cc @yokrysty @jborean93

### DIFF
--- a/src/Tmds.Ssh/BigIntegerExtensions.cs
+++ b/src/Tmds.Ssh/BigIntegerExtensions.cs
@@ -1,0 +1,22 @@
+using System.Numerics;
+using System.Diagnostics;
+
+namespace Tmds.Ssh;
+
+static class BigIntegerExtensions
+{
+    public static byte[] ToBEByteArray(this BigInteger integer, bool isUnsigned, int minLength)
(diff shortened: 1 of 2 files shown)


jborean93:
Just so I understand this was the same problem as the ECDSA keys I came across during implementation where we had to pad some values to get it working with .NET https://github.com/tmds/Tmds.Ssh/blob/7620058711e322bdba9a1a882b60a69c8253c106/src/Tmds.Ssh/PrivateKeyParser.OpenSsh.cs#L237?

tmds:
> Just so I understand this was the same problem as the ECDSA keys I came across during implementation where we had to pad some values to get it working with .NET 

Yes. .NET is having some length expectations when using these crypto classes, and both what we get from the SSH protocol, as what we get from BigInteger.ToByteArray is a minimal length representation. When the number needs less bytes in the minimal length representation, we need to padd it to meet the .NET crypto class expectation.

Issue #213, opened 2024-08-21
Support ssh-agent keys


tmds:
Support using private keys from ssh-agent

cc @jborean93

PR #281, opened 2024-12-19
Support auth using SSH Agent keys.

tmds:
Fixes https://github.com/tmds/Tmds.Ssh/issues/213.

cc @jborean93

### DIFF
--- a/src/Tmds.Ssh/SshAgent.cs
+++ b/src/Tmds.Ssh/SshAgent.cs
@@ -163,7 +163,7 @@ namespace Tmds.Ssh
             List<Identity> keys = new(capacity: nkeys);
             for (int i = 0; i < nkeys; i++)
             {
-                byte[] key_blob = reader.ReadStringAsBytes().ToArray();
+                byte[] key_blob = reader.ReadStringAsByteArray();
                 string comment = reader.ReadUtf8String();
                 keys.Add(new Identity() { Comment = comment, PublicKey = key_blob });
             }
@@ -180,7 +180,7 @@ namespace Tmds.Ssh
             var id = reader.ReadMessageId();
             if (id == SSH_AGENT_SIGN_RESPONSE)
             {
-                byte[] signature = reader.ReadStringAsBytes().ToArray();
+                byte[] signature = reader.ReadStringAsByteArray();
                 return signature;
             }
             else
--- a/src/Tmds.Ssh/SshAgent.cs
+++ b/src/Tmds.Ssh/SshAgent.cs
@@ -1,3 +1,6 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
 using System.Buffers;
 using System.Buffers.Binary;
 using System.Diagnostics;
--- a/src/Tmds.Ssh/SshAgent.cs
+++ b/src/Tmds.Ssh/SshAgent.cs
@@ -1,0 +1,279 @@
+using System.Buffers;
+using System.Buffers.Binary;
+using System.Diagnostics;
+using System.Net;
+using System.Net.Sockets;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Tmds.Ssh
+{
+    // https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent - SSH Agent Protocol
+    sealed class SshAgent : IDisposable
+    {
+        private const MessageId SSH_AGENTC_REQUEST_IDENTITIES = (MessageId)11;
+        private const MessageId SSH_AGENTC_SIGN_REQUEST = (MessageId)13;
+        private const MessageId SSH_AGENT_FAILURE = (MessageId)5;
+        private const MessageId SSH_AGENT_IDENTITIES_ANSWER = (MessageId)12;
+        private const MessageId SSH_AGENT_SIGN_RESPONSE = (MessageId)14;
+        private const uint SSH_AGENT_RSA_SHA2_256 = 2;
+        private const uint SSH_AGENT_RSA_SHA2_512 = 4;
+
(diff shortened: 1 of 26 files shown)


src/Tmds.Ssh/SshAgent.cs
```diff
@@ -0,0 +1,279 @@
...
+            public byte[] PublicKey { init; get; }
+        }
+
+        public static EndPoint? DefaultEndPoint
+        {
+            get
+            {
+                string? sshAuthSock = Environment.GetEnvironmentVariable("SSH_AUTH_SOCK");
+
+                if (OperatingSystem.IsWindows())
```
tmds, line 36: @jborean93 assuming you have a Windows machine/VM, is this something you'd want to look into?

src/Tmds.Ssh/UserAuthentication.SshAgentAuth.cs
```diff
@@ -0,0 +1,138 @@
...
+                if (context.IsSkipPublicAuthKey(key.PublicKey))
+                {
+                    continue;
+                }
+
+                SequenceReader reader = new SequenceReader(key.PublicKey);
+                Name keyType = reader.ReadName();
+
+                // Note: for unknown key type AlgorithmsForKeyType returns the key type as an algorithm. This is mostly true.
+                // This enables to use algorithms supported by the SSH Agent that this library doesn't directly support.
```
tmds, line 47: @jborean93 :point_up: this is nice, SSH Agent enables support for key types that aren't directly supported by the library.

src/Tmds.Ssh/SshAgent.cs
jborean93: Will try it out and get back to you.

jborean93:
Nice, I was planning on looking at this during the Christmas break but you've beaten me to it. I'll try out the changes on Windows to see what platform differences are there (if any).

src/Tmds.Ssh/SshAgent.cs
jborean93: On Windows it looks like it doesn't use a socket at all but rather a named pipe called `openssh-ssh-agent`. You can use [NamedPipeClientStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.pipes.namedpipeclientstream?view=net-9.0) to connect to it

```csharp
using var agentPipe = new NamedPipeClientStream("openssh-ssh-agent");
await agentPipe.ConnectAsync();
...
```

There's probably some considerations needed around

+ Whether any custom [PipeOption](https://learn.microsoft.com/en-us/dotnet/api/system.io.pipes.pipeoptions?view=net-9.0) is needed, like `Asynchronous`
+ Whether we should specify [TokenImpersonationLevel.Anonymous](https://learn.microsoft.com/en-us/dotnet/api/system.security.principal.tokenimpersonationlevel?view=net-9.0) or `Identification` to stop the service from pretending to be us
  + I'm not aware of any reason why it would need to do so but it's worth looking into

The pipe implements a `Stream` and not a `Socket` which will have some work needed to fit into the existing API model.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -0,0 +1,282 @@
...
+using System.Net.Sockets;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Tmds.Ssh
+{
+    // https://datatracker.ietf.org/doc/html/draft-miller-ssh-agent - SSH Agent Protocol
+    sealed class SshAgent : IDisposable
+    {
+        private const MessageId SSH_AGENTC_REQUEST_IDENTITIES = (MessageId)11;
```
jborean93, line 17: Is there a reason why you are using this as a const here rather than just adding it to the enum? I assume these ids are still unique even if it's specific to ssh-agent?

src/Tmds.Ssh/SshAgent.cs
tmds: > The pipe implements a Stream and not a Socket which will have some work needed to fit into the existing API model.

For this, you can change `SocketSshConnection` to accept a `Stream` instead of `Socket` and for the `Socket`-case create a `NetworkStream`. That changes will be useful for proxy jumping. Feel free to rename the class (e.g. `StreamSshConnection`.

src/Tmds.Ssh/SshAgent.cs
tmds: These ids are specific to SSH agent and overlap in value with the the regular SSH message ids.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -0,0 +1,282 @@
...
+        private const MessageId SSH_AGENTC_SIGN_REQUEST = (MessageId)13;
+        private const MessageId SSH_AGENT_FAILURE = (MessageId)5;
+        private const MessageId SSH_AGENT_IDENTITIES_ANSWER = (MessageId)12;
+        private const MessageId SSH_AGENT_SIGN_RESPONSE = (MessageId)14;
+        private const uint SSH_AGENT_RSA_SHA2_256 = 2;
+        private const uint SSH_AGENT_RSA_SHA2_512 = 4;
+
+        private static EndPoint? _defaultEndPoint;
```
tmds, line 25: @jborean93 for adding the named pipe support, we can probably change this from using an `EndPoint` to use a `string`. On Windows we will then assume this string is a named pipe address and on non-Windows it is a unix socket path.

tmds:
@jborean93 I've merged this and created some issues to tackle remaining things.

https://github.com/tmds/Tmds.Ssh/issues/288 is for the Windows support.

https://github.com/tmds/Tmds.Ssh/issues/284 is also important. I think you may be interested in working on this as well.

If you have an interest in working on any of the other issues, add a comment to it.

Issue #288, opened 2024-12-20
SSH Agent: support Windows


tmds:
https://github.com/tmds/Tmds.Ssh/pull/281 only includes detection for the SSH Agent on Unix. This is for adding Windows support.

cc @jborean93

jborean93:
I’ll try and tackle this one first. If it proves to be easy and simple I’ll have a look at the other ssh-agent issues after.

Issue #289, opened 2024-12-20
SSH Agent: add test


tmds:
Add tests that cover https://github.com/tmds/Tmds.Ssh/pull/281.

PR #291, opened 2024-12-21
Add ssh-agent support for Windows

jborean93:
Add support for using ssh-agent on Windows. The Win32 implementation of ssh-agent uses a named pipe rather than a unix domain socket on the listener but the underlying protocol is still the same.

Fixes https://github.com/tmds/Tmds.Ssh/issues/288


src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,60 +23,77 @@ sealed class SshAgent : IDisposable
...
+                {
+                    // There is no easy way to detect if a named pipe exists or
+                    // not. File.Exists is the fastest but will actually open a
+                    // connection and close it. Using this won't open the pipe
+                    // Directory.GetFiles("\\\\.\\pipe\\", endPoint).Length > 0
+                    // But it is slower.
+                    if (!File.Exists(@$"\\.\pipe\{_endPoint}"))
+                    {
+                        return false;
+                    }
```
jborean93, line 71: I'm happy to use the `Directory.GetFiles` method here if you prefer that but `File.Exists` is faster and the SSH agent implementation is able to handle multiple client connections. https://berserkerdotnet.github.io/blog/detecting-namedpipe-exists/ has some more background info for this problem.

If we don't do the check then `pipe.ConnectAsync` below will hang indefinitely as we don't have a timeout set. We should probably think about adding a default timeout as well to this call for both the pipe and socket connect calls.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,60 +23,77 @@ sealed class SshAgent : IDisposable
...
+
+                    pipe = new NamedPipeClientStream(
+                        ".",
```
jborean93, line 74: This is needed to refer to localhost, the overload for asynchronous and anon impersonation require the server name to be set.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,60 +23,77 @@ sealed class SshAgent : IDisposable
...
+                        _endPoint,
+                        PipeDirection.InOut,
+                        PipeOptions.Asynchronous,
+                        System.Security.Principal.TokenImpersonationLevel.Anonymous);
```
jborean93, line 78: I'm actually surprised this works, I need to dig into the ssh agent implementation for Windows to actually see what they are doing but this would indicate to me that the keys aren't protected per user but are machine wide which is a bit worrisome.

tmds:
Thanks for picking this up @jborean93! It's great to get some help for this.

src/Tmds.Ssh/SocketSshConnection.cs
```diff
@@ -1,250 +1,27 @@
...
-using System.Buffers;
-using System.Diagnostics;
 using System.Net.Sockets;
-using System.Text;
 using Microsoft.Extensions.Logging;
 
 namespace Tmds.Ssh;
 
-sealed class SocketSshConnection : SshConnection
+sealed class SocketSshConnection : StreamSshConnection
```
tmds, line 9: Can you remove this class completely? The NetworkStream that wraps the Socket can own it (for dispose).

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,60 +23,77 @@ sealed class SshAgent : IDisposable
...
-
-                return _defaultEndPoint;
+                return OperatingSystem.IsWindows()
+                    ? "openssh-ssh-agent"
+                    : Environment.GetEnvironmentVariable("SSH_AUTH_SOCK");
             }
         }
 
-        private readonly EndPoint _endPoint;
+        private readonly string _endPoint;
```
tmds, line 42: nit: rename endpoint to address.

src/Tmds.Ssh/SshAgent.cs
tmds: `File.Exists` is good as a quick check.

> If we don't do the check then pipe.ConnectAsync below will hang indefinitely as we don't have a timeout set. We should probably think about adding a default timeout as well to this call for both the pipe and socket connect calls.

Let's add a timeout.

```cs
// PipeConnectTImeOut = 1000; // 1s

using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
cts.CancelAfter(PipeConnectTImeOut);
...
pipe.ConnectAsync(cts.Token);
```

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,60 +23,77 @@ sealed class SshAgent : IDisposable
...
-        public SshAgent(EndPoint endPoint, SequencePool sequencePool)
+        public SshAgent(string endPoint, SequencePool sequencePool)
         {
             _endPoint = endPoint;
             _sequencePool = sequencePool;
         }
 
         public async ValueTask<bool> TryConnect(CancellationToken cancellationToken)
         {
+            NamedPipeClientStream? pipe = null;
```
tmds, line 55: After https://github.com/tmds/Tmds.Ssh/pull/291/files#r1895384096, try changing this to make the two branches produce a `Stream` (and move these declarations local to their branch), and then create the `StreamSshConnection` after the branches.

src/Tmds.Ssh/SshConnection.cs
```diff
@@ -22,6 +22,8 @@ protected SshConnection(SequencePool sequencePool)
 
     public abstract ValueTask<Packet> ReceivePacketAsync(CancellationToken ct, int maxLength = Constants.PreAuthMaxPacketLength);
     public abstract ValueTask SendPacketAsync(Packet packet, CancellationToken ct);
-    public abstract void Dispose();
     public abstract void SetEncryptorDecryptor(IPacketEncryptor packetEncoder, IPacketDecryptor packetDecoder, bool resetSequenceNumbers, bool throwIfReceiveSNZero);
+
+    public void Dispose() => Dispose(true);
+    protected virtual void Dispose(bool isDisposing) {}
```
tmds, line 28: I don't use this pattern unless there are native resources involved (in this type or derived ones).

src/Tmds.Ssh/SocketSshConnection.cs
jborean93: I'm not sure. The kept it as a sub class because I assumed we needed to keep the `Socket` object referenced somewhere. The socket stream is technically a separate object and I assumed that once the GC cleans up the Socket object it will close the stream as well but if that's not the case I can certainly simplify things.

tmds: `NetworkStream` will take ownership of the `Socket`, which means it will dispose the Socket when it gets disposed.

```
NetworkStream (System.Net.Sockets.Socket socket, bool ownsSocket);
```

As long as `StreamSshConnection` is considered reachable, the `Socket` is considered reachable too.

More technical:

The GC scans all the memory and follows locations it knows to be managed references. Anything not reachable from a stack is considered collectible. Objects that have a finalizer (like `~Socket` and `~SafeSocketHandle`) get finalized before being collected. If the user had called `Dispose` on these objects (for example through a `using StreamSshConnection` -> `NetworkStream.Dispose` -> `Socket.Dispose` -> `SafeSocketHandle.Dispose`) then that `Dispose` should included a call to `GC.SuppressFinalize` (on objects with a finalizer) which tells the GC it no longer needs to run the finalizer before reclaiming the memory on that object.

tmds:
@jborean93 when you continue working on this, please rebase so you use the latest changes when you try some things on Windows.

src/Tmds.Ssh/SshConnection.cs
jborean93: After the removal of the specialised Socket class this is no longer needed anyway but will keep it in mind for the future.

src/Tmds.Ssh/SshAgent.cs
jborean93: Done, let me know if you preferred what I've done in a different way.

src/Tmds.Ssh/SocketSshConnection.cs
jborean93: Thanks for the info, I was confusing this with [TcpClient.GetStream](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.tcpclient.getstream?view=net-9.0) where I was unsure who was the owner of the the actual underlying socket and what needed to be kept in referenced. I've changed it to rely on the `NetworkStream` ownership as you've mentioned here which tidies up the code quite nicely.

src/Tmds.Ssh/SshAgent.cs
jborean93: I just added the timeout to `ConnectAsync` rather than dealing with the token cancellation. I can change it to your method if you prefer that over the timeout overload.

src/Tmds.Ssh/SshAgent.cs
jborean93: I looked at the source code. Surprisingly it gets the PID of the client who has connected to the pipe and uses that to open the process token and get the user information that way.

https://github.com/PowerShell/openssh-portable/blob/0c3137f62115857cabb6520c2a06ec59e854a56e/contrib/win32/win32compat/ssh-agent/agent.c#L304-L310

This means we can connect with `Anonymous` as the server won't try to impersonate the client pipe thread but rather go through the PID.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -23,61 +23,77 @@ sealed class SshAgent : IDisposable
...
+                        PipeOptions.Asynchronous,
+                        System.Security.Principal.TokenImpersonationLevel.Anonymous);
+                    stream = pipe;
+
+                    await pipe.ConnectAsync(1000, cancellationToken).ConfigureAwait(false);
+                }
+                else
+                {
+                    Socket socket = new Socket(AddressFamily.Unix, SocketType.Stream, ProtocolType.Unspecified);
+                    stream = new NetworkStream(socket);
```
tmds, line 85: The default NetworkStream ctor doesn't own the Socket, you need to call the overload that accepts a boolean.

src/Tmds.Ssh/SshAgent.cs
tmds: I didn't know that overload exists. This does exactly what we need.

src/Tmds.Ssh/SshAgent.cs
jborean93: Ah my bad, I assumed it was the default, has been updated.

tmds:
Thanks for making this work @jborean93!

I'll look into adding a test for ssh-agent auth (https://github.com/tmds/Tmds.Ssh/issues/289), and then I'll make a release.

jborean93:
Thanks for reviewing and sharing info like the ownership side.

tmds:
This is part of the 0.10.0 version which was just uploaded to nuget.org.

PR #306, opened 2024-12-28
Add support for curve25519-sha256 key exchange method

scott-xu:
This PR adds support for `curve25519-sha2` key exchange method described by https://datatracker.ietf.org/doc/html/rfc8731

Close https://github.com/tmds/Tmds.Ssh/issues/307

### DIFF
--- a/README.md
+++ b/README.md
@@ -566,6 +566,8 @@ Supported server key algorithms:
 - rsa-sha2-256
 
 Supported key exchange methods:
+- curve25519-sha256
+- curve25519-sha256@libssh.org
 - ecdh-sha2-nistp256
 - ecdh-sha2-nistp384
 - ecdh-sha2-nistp521
--- a/README.md
+++ b/README.md
@@ -566,8 +566,7 @@ Supported server key algorithms:
 - rsa-sha2-256
 
 Supported key exchange methods:
-- curve25519-sha256
-- curve25519-sha256@libssh.org
+- curve25519-sha256, curve25519-sha256@libssh.org
 - ecdh-sha2-nistp256
 - ecdh-sha2-nistp384
 - ecdh-sha2-nistp521
--- a/src/Tmds.Ssh/Curve25519KeyExchange.cs
+++ b/src/Tmds.Ssh/Curve25519KeyExchange.cs
@@ -3,7 +3,6 @@
 
 using System.Buffers;
 using System.Security.Cryptography;
-using System.Numerics;
 using Microsoft.Extensions.Logging;
 using Org.BouncyCastle.Crypto.Generators;
 using Org.BouncyCastle.Crypto.Parameters;
@@ -44,7 +43,7 @@ class Curve25519KeyExchange : KeyExchange
         PublicKey publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
 
         // Compute shared secret.
-        BigInteger sharedSecret;
+        byte[] sharedSecret;
         try
         {
             sharedSecret = DeriveSharedSecret(x25519KeyPair.Private, new X25519PublicKeyParameters(ecdhReply.q_s));
@@ -63,7 +62,7 @@ class Curve25519KeyExchange : KeyExchange
         return CalculateKeyExchangeOutput(input, sequencePool, sharedSecret, exchangeHash, _hashAlgorithmName);
     }
 
-    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, byte[] q_c, byte[] q_s, BigInteger sharedSecret, HashAlgorithmName hashAlgorithmName)
+    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, byte[] q_c, byte[] q_s, byte[] sharedSecret, HashAlgorithmName hashAlgorithmName)
     {
         /*
             string   V_C, client's identification string (CR and LF excluded)
@@ -84,7 +83,7 @@ class Curve25519KeyExchange : KeyExchange
         writer.WriteString(public_host_key);
         writer.WriteString(q_c);
         writer.WriteString(q_s);
-        writer.WriteMPInt(sharedSecret);
+        writer.WriteString(sharedSecret);
 
         using IncrementalHash hash = IncrementalHash.CreateHash(hashAlgorithmName);
         foreach (var segment in sequence.AsReadOnlySequence())
@@ -94,7 +93,7 @@ class Curve25519KeyExchange : KeyExchange
         return hash.GetHashAndReset();
     }
 
-    private static BigInteger DeriveSharedSecret(AsymmetricKeyParameter privateKey, AsymmetricKeyParameter peerPublicKey)
+    private static byte[] DeriveSharedSecret(AsymmetricKeyParameter privateKey, AsymmetricKeyParameter peerPublicKey)
     {
         var keyAgreement = new X25519Agreement();
         keyAgreement.Init(privateKey);
@@ -103,7 +102,7 @@ class Curve25519KeyExchange : KeyExchange
         keyAgreement.CalculateAgreement(peerPublicKey, rawSecretAgreement);
         var sharedSecret = rawSecretAgreement.ToBigInteger();
         rawSecretAgreement.AsSpan().Clear();
-        return sharedSecret;
+        return sharedSecret.ToMPIntByteArray();
     }
 
     private static Packet CreateEcdhInitMessage(SequencePool sequencePool, ReadOnlySpan<byte> q_c)
--- a/src/Tmds.Ssh/Curve25519KeyExchange.cs
+++ b/src/Tmds.Ssh/Curve25519KeyExchange.cs
@@ -15,14 +15,11 @@ using Org.BouncyCastle.Crypto.Agreement;
 namespace Tmds.Ssh;
 
 // Curve25519 Key Exchange: https://datatracker.ietf.org/doc/html/rfc8731
-class Curve25519KeyExchange : IKeyExchangeAlgorithm
+class Curve25519KeyExchange : KeyExchange, IKeyExchangeAlgorithm
 {
-    private readonly HashAlgorithmName _hashAlgorithmName;
-
-    public Curve25519KeyExchange(HashAlgorithmName hashAlgorithmName)
-    {
-        _hashAlgorithmName = hashAlgorithmName;
-    }
+    public Curve25519KeyExchange()
+        :base(HashAlgorithmName.SHA256)
+    { }
 
     public async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContext context, IHostKeyVerification hostKeyVerification, Packet firstPacket, KeyExchangeInput input, ILogger logger, CancellationToken ct)
     {
@@ -46,14 +43,7 @@ class Curve25519KeyExchange : IKeyExchangeAlgorithm
         var ecdhReply = ParceEcdhReply(ecdhReplyMsg);
 
         // Verify received key is valid.
-        connectionInfo.ServerKey = new HostKey(ecdhReply.public_host_key);
-        await hostKeyVerification.VerifyAsync(connectionInfo, ct).ConfigureAwait(false);
-
-        var publicHostKey = PublicKey.CreateFromSshKey(ecdhReply.public_host_key);
-        if (publicHostKey is RsaPublicKey rsaPublicKey && rsaPublicKey.KeySize < input.MinimumRSAKeySize)
-        {
-            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, $"Server RSA key size {rsaPublicKey.KeySize} is less than {input.MinimumRSAKeySize}.", connectionInfo);
-        }
+        var publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
 
         // Compute shared secret.
         BigInteger sharedSecret;
@@ -70,22 +60,9 @@ class Curve25519KeyExchange : IKeyExchangeAlgorithm
         byte[] exchangeHash = CalculateExchangeHash(sequencePool, input.ConnectionInfo, input.ClientKexInitMsg, input.ServerKexInitMsg, ecdhReply.public_host_key.Data, q_c, ecdhReply.q_s, sharedSecret);
 
         // Verify the server's signature.
-        if (!publicHostKey.VerifySignature(input.HostKeyAlgorithms, exchangeHash, ecdhReply.exchange_hash_signature))
-        {
-            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, "Signature does not match host key.", connectionInfo);
-        }
+        VerifySignature(publicHostKey, input.HostKeyAlgorithms, exchangeHash, ecdhReply.exchange_hash_signature, connectionInfo);
 
-        byte[] sessionId = input.ConnectionInfo.SessionId ?? exchangeHash;
-        byte[] initialIVC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'A', sessionId, input.InitialIVC2SLength);
-        byte[] initialIVS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'B', sessionId, input.InitialIVS2CLength);
-        byte[] encryptionKeyC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'C', sessionId, input.EncryptionKeyC2SLength);
-        byte[] encryptionKeyS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'D', sessionId, input.EncryptionKeyS2CLength);
-        byte[] integrityKeyC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'E', sessionId, input.IntegrityKeyC2SLength);
-        byte[] integrityKeyS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'F', sessionId, input.IntegrityKeyS2CLength);
-
-        return new KeyExchangeOutput(exchangeHash,
-            initialIVS2C, encryptionKeyS2C, integrityKeyS2C,
-            initialIVC2S, encryptionKeyC2S, integrityKeyC2S);
+        return CalculateKeyExchangeOutput(input, sequencePool, sharedSecret, exchangeHash);
     }
 
     private byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, byte[] q_c, byte[] q_s, BigInteger sharedSecret)
@@ -117,58 +94,6 @@ class Curve25519KeyExchange : IKeyExchangeAlgorithm
             hash.AppendData(segment.Span);
         }
         return hash.GetHashAndReset();
-    }
-
-    private byte[] CalculateKey(SequencePool sequencePool, BigInteger sharedSecret, byte[] exchangeHash, byte c, byte[] sessionId, int keyLength)
-    {
-        // https://tools.ietf.org/html/rfc4253#section-7.2
-
-        byte[] key = new byte[keyLength];
-        int keyOffset = 0;
-
-        // HASH(K || H || c || session_id)
-        using Sequence sequence = sequencePool.RentSequence();
-        var writer = new SequenceWriter(sequence);
-        writer.WriteMPInt(sharedSecret);
-        writer.Write(exchangeHash);
-        writer.WriteByte(c);
-        writer.Write(sessionId);
-
-        using IncrementalHash hash = IncrementalHash.CreateHash(_hashAlgorithmName);
-        foreach (var segment in sequence.AsReadOnlySequence())
-        {
-            hash.AppendData(segment.Span);
-        }
-        byte[] K1 = hash.GetHashAndReset();
-        Append(key, K1, ref keyOffset);
-
-        while (keyOffset != key.Length)
-        {
-            sequence.Clear();
-
-            // K3 = HASH(K || H || K1 || K2)
-            writer = new SequenceWriter(sequence);
-            writer.WriteMPInt(sharedSecret);
-            writer.Write(exchangeHash);
-            writer.Write(key.AsSpan(0, keyOffset));
-
-            foreach (var segment in sequence.AsReadOnlySequence())
-            {
-                hash.AppendData(segment.Span);
-            }
-            byte[] Kn = hash.GetHashAndReset();
-
-            Append(key, Kn, ref keyOffset);
-        }
-
-        return key;
-
-        static void Append(byte[] key, byte[] append, ref int offset)
-        {
-            int available = Math.Min(append.Length, key.Length - offset);
-            append.AsSpan().Slice(0, available).CopyTo(key.AsSpan(offset));
-            offset += available;
-        }
     }
 
     private BigInteger DeriveSharedSecret(AsymmetricKeyParameter privateKey, AsymmetricKeyParameter peerPublicKey)
--- a/src/Tmds.Ssh/Curve25519KeyExchange.cs
+++ b/src/Tmds.Ssh/Curve25519KeyExchange.cs
@@ -15,13 +15,11 @@ using Org.BouncyCastle.Crypto.Agreement;
 namespace Tmds.Ssh;
 
 // Curve25519 Key Exchange: https://datatracker.ietf.org/doc/html/rfc8731
-class Curve25519KeyExchange : KeyExchange, IKeyExchangeAlgorithm
+class Curve25519KeyExchange : KeyExchange
 {
-    public Curve25519KeyExchange()
-        :base(HashAlgorithmName.SHA256)
-    { }
+    private readonly HashAlgorithmName _hashAlgorithmName = HashAlgorithmName.SHA256;
 
-    public async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContext context, IHostKeyVerification hostKeyVerification, Packet firstPacket, KeyExchangeInput input, ILogger logger, CancellationToken ct)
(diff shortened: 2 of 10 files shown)


scott-xu:
The CI action is awaiting approval @tmds

tmds:
@scott-xu looks good! I'd like to share some code with the `ECDHKeyExchange` class by adding a base `KeyExchange` class. I'm going to highlight the regions of code that I think can be in the base class.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        // Verify received key is valid.
+        connectionInfo.ServerKey = new HostKey(ecdhReply.public_host_key);
+        await hostKeyVerification.VerifyAsync(connectionInfo, ct).ConfigureAwait(false);
+
+        var publicHostKey = PublicKey.CreateFromSshKey(ecdhReply.public_host_key);
+        if (publicHostKey is RsaPublicKey rsaPublicKey && rsaPublicKey.KeySize < input.MinimumRSAKeySize)
+        {
+            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, $"Server RSA key size {rsaPublicKey.KeySize} is less than {input.MinimumRSAKeySize}.", connectionInfo);
+        }
```
tmds, line 56:
```
PublicKey publicHostKey = await VerifyHostKeyAsync(...).ConfigureAwait(false);
```
using `VerifyHostKeyAsync` from base class.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,215 @@
...
+
+        // Generate exchange hash.
+        byte[] exchangeHash = CalculateExchangeHash(sequencePool, input.ConnectionInfo, input.ClientKexInitMsg, input.ServerKexInitMsg, ecdhReply.public_host_key.Data, q_c, ecdhReply.q_s, sharedSecret);
+
+        // Verify the server's signature.
+        if (!publicHostKey.VerifySignature(input.HostKeyAlgorithms, exchangeHash, ecdhReply.exchange_hash_signature))
+        {
+            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, "Signature does not match host key.", connectionInfo);
+        }
```
tmds, line 76:
```
VerifySignature(publicHostKey, ...);
```
using `VerifySignature` from base class.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,215 @@
...
+        byte[] initialIVC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'A', sessionId, input.InitialIVC2SLength);
+        byte[] initialIVS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'B', sessionId, input.InitialIVS2CLength);
+        byte[] encryptionKeyC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'C', sessionId, input.EncryptionKeyC2SLength);
+        byte[] encryptionKeyS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'D', sessionId, input.EncryptionKeyS2CLength);
+        byte[] integrityKeyC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'E', sessionId, input.IntegrityKeyC2SLength);
+        byte[] integrityKeyS2C = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'F', sessionId, input.IntegrityKeyS2CLength);
+
+        return new KeyExchangeOutput(exchangeHash,
+            initialIVS2C, encryptionKeyS2C, integrityKeyS2C,
+            initialIVC2S, encryptionKeyC2S, integrityKeyC2S);
```
tmds, line 88:
```
return new CalculateKeyExchangeOutput(...);
```
using `CalculateKeyExchangeOutput` from base class.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,215 @@
...
+using Org.BouncyCastle.Crypto.Agreement;
+
+namespace Tmds.Ssh;
+
+// Curve25519 Key Exchange: https://datatracker.ietf.org/doc/html/rfc8731
+class Curve25519KeyExchange : IKeyExchangeAlgorithm
+{
+    private readonly HashAlgorithmName _hashAlgorithmName;
+
+    public Curve25519KeyExchange(HashAlgorithmName hashAlgorithmName)
```
tmds, line 22: nit: remote the constructor argument, as we're never setting it to anything other than `HashAlgorithmName.SHA256`.

scott-xu:
That's exactly what I'm doing right now and is the reason why I convert it back to *Draft*.

src/Tmds.Ssh/Curve25519KeyExchange.cs
scott-xu: Note: the `sharedSecret` is `mpint` when ECDH and Curve25519, but is octet string when hybrid. See https://www.ietf.org/archive/id/draft-ietf-sshm-ntruprime-ssh-01.html

scott-xu:
I got some challenges when I try to put most logics in the base `KeyExchange` class and let derived class implement the difference.
For example, I tried to abstract `ReadServerExchangeValue` method and put the implementation in derived class. However it doesn't work because below `reader` is `ref struct`.
```C#
var reader = packet.GetReader();
reader.ReadMessageId(MessageId.SSH_MSG_KEX_ECDH_REPLY);
SshKey public_host_key = reader.ReadSshKey();
ReadServerExchangeValue(reader);
ReadOnlySequence<byte> exchange_hash_signature = reader.ReadStringAsBytes();
reader.ReadEnd();
return (
    public_host_key,
    exchange_hash_signature);
```

```C#
protected override void WriteServerExchangeValue(SequenceWriter writer)
{
    writer.WriteString(q_s!);
}
```

src/Tmds.Ssh/Curve25519KeyExchange.cs
tmds: We can change `CalculateKey` to take the shared secret as a `ReadOnlySpan<byte>`.

In `BigIntegerExtensions` we can add something like:

```cs
    public static byte[] ToMPIntString(this BigInteger integer)
        => integer == BigInteger.Zero ? Array.Empty<byte>() : integer.ToByteArray(isUnsigned: false, isBigEndian: true);
```

and call it here to change the `BigInteger` into an octet string.

tmds: `ToMPIntString` -> `ToMPIntByteArray` is a better name. (string in the method name may be confusing)

tmds:
> I got some challenges when I try to put most logics in the base KeyExchange class and let derived class implement the difference.

I don't want to try and split the key exchange in several abstract methods that a derived class then implements because of the challenges.

I would like to focus on the sections of code I commented on and let each derived class use them in their `TryExchangeAsync` implementation by calling the base class method.

src/Tmds.Ssh/Curve25519KeyExchange.cs
scott-xu: > and call it here to change the `BigInteger` into an octet string.

Sorry I'm not quite understand. Regarding hybrid kex, the derived secret is octet string. Where does `BigInteger` come from?

tmds: If we'd like to use `CalculateKeyExchangeOutput` in the hybrid kex, we want to pass it an octet string. So we can change the implementation (in `ECDHKeyExchange` and `Curve25519KeyExchange`) to convert the `BigInteger` shared secret into a byte array (the octet string content) that gets passed to `CalculateKeyExchangeOutput`.

src/Tmds.Ssh/KeyExchange.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Numerics;
+using System.Security.Cryptography;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Tmds.Ssh;
+
+abstract class KeyExchange
```
tmds, line 12: Please put the `IKeyExchangeAlgorithm` here and remove it from the derived classes.

README.md
```diff
@@ -566,6 +566,8 @@ Supported server key algorithms:
 - rsa-sha2-256
 
 Supported key exchange methods:
+- curve25519-sha256
+- curve25519-sha256@libssh.org
```
tmds, line 570: nit: put these on a single line (because they are the same algorithm).

src/Tmds.Ssh/KeyExchange.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    protected readonly HashAlgorithmName _hashAlgorithmName;
```
tmds, line 14: Please remove the field from the base class and let the derived class track it.

src/Tmds.Ssh/KeyExchange.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    public static void VerifySignature(PublicKey publicHostKey, IReadOnlyList<Name> allowedAlgorithms, byte[] exchangeHash, ReadOnlySequence<byte> exchange_hash_signature, SshConnectionInfo connectionInfo)
+    {
+        if (!publicHostKey.VerifySignature(allowedAlgorithms, exchangeHash, exchange_hash_signature))
+        {
+            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, "Signature does not match host key.", connectionInfo);
+        }
+    }
+
+    public KeyExchangeOutput CalculateKeyExchangeOutput(KeyExchangeInput input, SequencePool sequencePool, BigInteger sharedSecret, byte[] exchangeHash)
```
tmds, line 44: Can this be `static`?

src/Tmds.Ssh/KeyExchange.cs
```diff
@@ -0,0 +1,111 @@
...
+abstract class KeyExchange
+{
+    protected readonly HashAlgorithmName _hashAlgorithmName;
+
+    public KeyExchange(HashAlgorithmName hashAlgorithmName)
+    {
+        this._hashAlgorithmName = hashAlgorithmName;
+    }
+
+    public static async Task<PublicKey> VerifyHostKeyAsync(IHostKeyVerification hostKeyVerification, KeyExchangeInput input, SshKey public_host_key, CancellationToken ct)
```
tmds, line 21: `public` -> `protected` (same for the other methods).

src/Tmds.Ssh/Curve25519KeyExchange.cs
tmds: The change for this is:
```diff
diff --git a/src/Tmds.Ssh/BigIntegerExtensions.cs b/src/Tmds.Ssh/BigIntegerExtensions.cs
index 98124f2..5c189a3 100644
--- a/src/Tmds.Ssh/BigIntegerExtensions.cs
+++ b/src/Tmds.Ssh/BigIntegerExtensions.cs
@@ -19,4 +19,7 @@ public static byte[] ToBEByteArray(this BigInteger integer, bool isUnsigned, int
         }
         return array;
     }
+
+    public static byte[] ToMPIntByteArray(this BigInteger integer)
+        => integer == BigInteger.Zero ? Array.Empty<byte>() : integer.ToByteArray(isUnsigned: false, isBigEndian: true);
 }
\ No newline at end of file
diff --git a/src/Tmds.Ssh/Curve25519KeyExchange.cs b/src/Tmds.Ssh/Curve25519KeyExchange.cs
index 4b74ce1..b60a4d2 100644
--- a/src/Tmds.Ssh/Curve25519KeyExchange.cs
+++ b/src/Tmds.Ssh/Curve25519KeyExchange.cs
@@ -44,7 +44,7 @@ public override async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContex
         var publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
 
         // Compute shared secret.
-        BigInteger sharedSecret;
+        byte[] sharedSecret;
         try
         {
             sharedSecret = DeriveSharedSecret(x25519KeyPair.Private, new X25519PublicKeyParameters(ecdhReply.q_s));
@@ -63,7 +63,7 @@ public override async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContex
         return CalculateKeyExchangeOutput(input, sequencePool, sharedSecret, exchangeHash, _hashAlgorithmName);
     }
 
-    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, byte[] q_c, byte[] q_s, BigInteger sharedSecret, HashAlgorithmName hashAlgorithmName)
+    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, byte[] q_c, byte[] q_s, byte[] sharedSecret, HashAlgorithmName hashAlgorithmName)
     {
         /*
             string   V_C, client's identification string (CR and LF excluded)
@@ -84,7 +84,7 @@ private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnec
         writer.WriteString(public_host_key);
         writer.WriteString(q_c);
         writer.WriteString(q_s);
-        writer.WriteMPInt(sharedSecret);
+        writer.WriteString(sharedSecret);
 
         using IncrementalHash hash = IncrementalHash.CreateHash(hashAlgorithmName);
         foreach (var segment in sequence.AsReadOnlySequence())
@@ -94,7 +94,7 @@ private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnec
         return hash.GetHashAndReset();
     }
 
-    private static BigInteger DeriveSharedSecret(AsymmetricKeyParameter privateKey, AsymmetricKeyParameter peerPublicKey)
+    private static byte[] DeriveSharedSecret(AsymmetricKeyParameter privateKey, AsymmetricKeyParameter peerPublicKey)
     {
         var keyAgreement = new X25519Agreement();
         keyAgreement.Init(privateKey);
@@ -103,7 +103,7 @@ private static BigInteger DeriveSharedSecret(AsymmetricKeyParameter privateKey,
         keyAgreement.CalculateAgreement(peerPublicKey, rawSecretAgreement);
         var sharedSecret = rawSecretAgreement.ToBigInteger();
         rawSecretAgreement.AsSpan().Clear();
-        return sharedSecret;
+        return sharedSecret.ToMPIntByteArray();
     }
 
     private static Packet CreateEcdhInitMessage(SequencePool sequencePool, ReadOnlySpan<byte> q_c)
diff --git a/src/Tmds.Ssh/ECDHKeyExchange.cs b/src/Tmds.Ssh/ECDHKeyExchange.cs
index 21d3913..47bf2b9 100644
--- a/src/Tmds.Ssh/ECDHKeyExchange.cs
+++ b/src/Tmds.Ssh/ECDHKeyExchange.cs
@@ -39,7 +39,7 @@ public override async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContex
         var publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
 
         // Compute shared secret.
-        BigInteger sharedSecret;
+        byte[] sharedSecret;
         try
         {
             sharedSecret = DeriveSharedSecret(ecdh, _ecCurve, ecdhReply.q_s);
@@ -58,7 +58,7 @@ public override async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContex
         return CalculateKeyExchangeOutput(input, sequencePool, sharedSecret, exchangeHash, _hashAlgorithmName);
     }
 
-    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, ECPoint q_c, ECPoint q_s, BigInteger sharedSecret, HashAlgorithmName hashAlgorithmName)
+    private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInfo connectionInfo, ReadOnlyPacket clientKexInitMsg, ReadOnlyPacket serverKexInitMsg, byte[] public_host_key, ECPoint q_c, ECPoint q_s, byte[] sharedSecret, HashAlgorithmName hashAlgorithmName)
     {
         /*
             string   V_C, client's identification string (CR and LF excluded)
@@ -79,7 +79,7 @@ private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnec
         writer.WriteString(public_host_key);
         writer.WriteString(q_c);
         writer.WriteString(q_s);
-        writer.WriteMPInt(sharedSecret);
+        writer.WriteString(sharedSecret);
 
         using IncrementalHash hash = IncrementalHash.CreateHash(hashAlgorithmName);
         foreach (var segment in sequence.AsReadOnlySequence())
@@ -89,7 +89,7 @@ private static byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnec
         return hash.GetHashAndReset();
     }
 
-    private static BigInteger DeriveSharedSecret(ECDiffieHellman ecdh, ECCurve curve, ECPoint q)
+    private static byte[] DeriveSharedSecret(ECDiffieHellman ecdh, ECCurve curve, ECPoint q)
     {
         ECParameters parameters = new ECParameters
         {
@@ -101,7 +101,7 @@ private static BigInteger DeriveSharedSecret(ECDiffieHellman ecdh, ECCurve curve
         byte[] rawSecretAgreement = ecdh.DeriveRawSecretAgreement(peerPublicKey);
         var sharedSecret = rawSecretAgreement.ToBigInteger();
         rawSecretAgreement.AsSpan().Clear();
-        return sharedSecret;
+        return sharedSecret.ToMPIntByteArray();
     }
 
     private static Packet CreateEcdhInitMessage(SequencePool sequencePool, ECPoint q_c)
diff --git a/src/Tmds.Ssh/KeyExchange.cs b/src/Tmds.Ssh/KeyExchange.cs
index 25408b8..9ceab22 100644
--- a/src/Tmds.Ssh/KeyExchange.cs
+++ b/src/Tmds.Ssh/KeyExchange.cs
@@ -32,7 +32,7 @@ protected static void VerifySignature(PublicKey publicHostKey, IReadOnlyList<Nam
         }
     }
 
-    protected static KeyExchangeOutput CalculateKeyExchangeOutput(KeyExchangeInput input, SequencePool sequencePool, BigInteger sharedSecret, byte[] exchangeHash, HashAlgorithmName hashAlgorithmName)
+    protected static KeyExchangeOutput CalculateKeyExchangeOutput(KeyExchangeInput input, SequencePool sequencePool, byte[] sharedSecret, byte[] exchangeHash, HashAlgorithmName hashAlgorithmName)
     {
         byte[] sessionId = input.ConnectionInfo.SessionId ?? exchangeHash;
         byte[] initialIVC2S = CalculateKey(sequencePool, sharedSecret, exchangeHash, (byte)'A', sessionId, input.InitialIVC2SLength, hashAlgorithmName);
@@ -47,7 +47,7 @@ protected static KeyExchangeOutput CalculateKeyExchangeOutput(KeyExchangeInput i
             initialIVC2S, encryptionKeyC2S, integrityKeyC2S);
     }
 
-    protected static byte[] CalculateKey(SequencePool sequencePool, BigInteger sharedSecret, byte[] exchangeHash, byte c, byte[] sessionId, int keyLength, HashAlgorithmName hashAlgorithmName)
+    protected static byte[] CalculateKey(SequencePool sequencePool, byte[] sharedSecret, byte[] exchangeHash, byte c, byte[] sessionId, int keyLength, HashAlgorithmName hashAlgorithmName)
     {
         // https://tools.ietf.org/html/rfc4253#section-7.2
 
@@ -57,7 +57,7 @@ protected static byte[] CalculateKey(SequencePool sequencePool, BigInteger share
         // HASH(K || H || c || session_id)
         using Sequence sequence = sequencePool.RentSequence();
         var writer = new SequenceWriter(sequence);
-        writer.WriteMPInt(sharedSecret);
+        writer.WriteString(sharedSecret);
         writer.Write(exchangeHash);
         writer.WriteByte(c);
         writer.Write(sessionId);
@@ -76,7 +76,7 @@ protected static byte[] CalculateKey(SequencePool sequencePool, BigInteger share
 
             // K3 = HASH(K || H || K1 || K2)
             writer = new SequenceWriter(sequence);
-            writer.WriteMPInt(sharedSecret);
+            writer.WriteString(sharedSecret);
             writer.Write(exchangeHash);
             writer.Write(key.AsSpan(0, keyOffset));

```

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,135 @@
...
+{
+    private readonly HashAlgorithmName _hashAlgorithmName = HashAlgorithmName.SHA256;
+
+    public override async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContext context, IHostKeyVerification hostKeyVerification, Packet firstPacket, KeyExchangeInput input, ILogger logger, CancellationToken ct)
+    {
+        var sequencePool = context.SequencePool;
+        var connectionInfo = input.ConnectionInfo;
+
+        AsymmetricCipherKeyPair x25519KeyPair;
+        using (var _randomGenerator = new CryptoApiRandomGenerator())
```
tmds, line 28: nit: drop the `_` in `_randomGenerator`.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            var x25519KeyPairGenerator = new X25519KeyPairGenerator();
+            x25519KeyPairGenerator.Init(new X25519KeyGenerationParameters(new SecureRandom(_randomGenerator)));
+            x25519KeyPair = x25519KeyPairGenerator.GenerateKeyPair();
+        }
+
+        // Send ECDH_INIT.
+        var q_c = ((X25519PublicKeyParameters)x25519KeyPair.Public).GetEncoded();
```
tmds, line 36: `var` -> `byte[]`

The code style only uses `var` when the right hand side is a constructor or explicitly contains the returned type as part of the member name being called.

src/Tmds.Ssh/Curve25519KeyExchange.cs
```diff
@@ -0,0 +1,135 @@
...
+        await context.SendPacketAsync(CreateEcdhInitMessage(sequencePool, q_c), ct).ConfigureAwait(false);
+
+        // Receive ECDH_REPLY.
+        using Packet ecdhReplyMsg = await context.ReceivePacketAsync(MessageId.SSH_MSG_KEX_ECDH_REPLY, firstPacket.Move(), ct).ConfigureAwait(false);
+        var ecdhReply = ParceEcdhReply(ecdhReplyMsg);
+
+        // Verify received key is valid.
+        var publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
```
tmds, line 44: `var` -> `PublicKey`

src/Tmds.Ssh/ECDHKeyExchange.cs
```diff
@@ -9,18 +9,18 @@
...
+class ECDHKeyExchange : KeyExchange
 {
     private readonly ECCurve _ecCurve;
     private readonly HashAlgorithmName _hashAlgorithmName;
 
     public ECDHKeyExchange(ECCurve ecCurve, HashAlgorithmName hashAlgorithmName)
     {
         _ecCurve = ecCurve;
-        _hashAlgorithmName = hashAlgorithmName;
+        this._hashAlgorithmName = hashAlgorithmName;
```
tmds, line 20: remove the `this.`.

src/Tmds.Ssh/ECDHKeyExchange.cs
```diff
@@ -36,49 +36,29 @@ public async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContext context
...
         // Verify received key is valid.
-        connectionInfo.ServerKey = new HostKey(ecdhReply.public_host_key);
-        await hostKeyVerification.VerifyAsync(connectionInfo, ct).ConfigureAwait(false);
-
-        var publicHostKey = PublicKey.CreateFromSshKey(ecdhReply.public_host_key);
-        if (publicHostKey is RsaPublicKey rsaPublicKey && rsaPublicKey.KeySize < input.MinimumRSAKeySize)
-        {
-            throw new ConnectFailedException(ConnectFailedReason.KeyExchangeFailed, $"Server RSA key size {rsaPublicKey.KeySize} is less than {input.MinimumRSAKeySize}.", connectionInfo);
-        }
+        var publicHostKey = await VerifyHostKeyAsync(hostKeyVerification, input, ecdhReply.public_host_key, ct).ConfigureAwait(false);
```
tmds, line 39: `var` -> `PublicKey`.

scott-xu:
Might be good to use another PR for `sntrup761x25519sha512` and `mlkem768x25519sha256`

src/Tmds.Ssh/Curve25519KeyExchange.cs
scott-xu: Do you want me to apply the change in this PR or the upcoming hybrid kex PR?

tmds: Let's include it here, so what is provided from the base class is more "generic".

scott-xu: Done

src/Tmds.Ssh/ECDHKeyExchange.cs
```diff
@@ -99,86 +78,31 @@ private byte[] CalculateExchangeHash(SequencePool sequencePool, SshConnectionInf
...
+        writer.WriteString(sharedSecret);
 
-        using IncrementalHash hash = IncrementalHash.CreateHash(_hashAlgorithmName);
+        using IncrementalHash hash = IncrementalHash.CreateHash(hashAlgorithmName);
         foreach (var segment in sequence.AsReadOnlySequence())
         {
             hash.AppendData(segment.Span);
         }
         return hash.GetHashAndReset();
     }
```
scott-xu, line 89: Looks like `CalculateExchangeHash` can be shared as well

tmds: It could with some work that deals with the `ECPoint`. I'm going to merge this PR as it is now. We can tackle this in a separate PR.

scott-xu: 👍🏼

src/Tmds.Ssh/Curve25519KeyExchange.cs
scott-xu: Do you want me rename the file as `Curve25519Sha256KeyExchange.cs`?

scott-xu: Renamed in https://github.com/tmds/Tmds.Ssh/pull/311

Issue #302, opened 2024-12-27
Add support for hybrid key exchange methods


scott-xu:
Suggest adding support for `sntrup761x25519-sha512` and `mlkem768x25519-sha256` key exchange method which is described at https://www.ietf.org/archive/id/draft-ietf-sshm-ntruprime-ssh-01.html and https://www.ietf.org/archive/id/draft-kampanakis-curdle-ssh-pq-ke-05.html

I created 2 PRs in SSH.NET. See https://github.com/sshnet/SSH.NET/pull/1562 and https://github.com/sshnet/SSH.NET/pull/1563

@tmds Let me know if you want me port the two PRs to this repository.

tmds:
@scott-xu it would be very interesting to have support for this, and thank you for offering to implement it!

 I'm not sure if the abstractions we have are fit to plug this in. We'll figure that out as we go.
Let's start with `sntrup761x25519-sha512`, and tackle `mlkem768x25519-sha256` after that.

For tests, please add something like https://github.com/tmds/Tmds.Ssh/blob/main/test/Tmds.Ssh.Tests/HostKeyAlgorithmTests.cs but for setting `KeyExchangeAlgorithms` instead of `ServerHostKeyAlgorithms`.

I'm finishing off some things to make a December release, and this can be part of the January release.

scott-xu:
Let me try `curve25519-sha256` first.

PR #311, opened 2024-12-31
Add support for mlkem768x25519-sha256 key exchange method

scott-xu:
Close https://github.com/tmds/Tmds.Ssh/issues/302


src/Tmds.Ssh/Curve25519Sha256KeyExchange.cs
```diff
@@ -14,7 +14,7 @@
 namespace Tmds.Ssh;
 
 // Curve25519 Key Exchange: https://datatracker.ietf.org/doc/html/rfc8731
-class Curve25519KeyExchange : KeyExchange
+class Curve25519Sha256KeyExchange : KeyExchange
```
tmds, line 17: Can you keep the old name and rename the file instead?

src/Tmds.Ssh/MLKem768X25519Sha256KeyExchange.cs
```diff
@@ -0,0 +1,122 @@
...
+using Org.BouncyCastle.Crypto.Kems;
+using Org.BouncyCastle.Crypto.Parameters;
+using Org.BouncyCastle.Crypto.Prng;
+using Org.BouncyCastle.Security;
+
+namespace Tmds.Ssh;
+
+// Post-Quantum Traditional (PQ/T) Hybrid key exchange methods using ML-KEM-768 and X25519 with SHA-256
+// https://www.ietf.org/archive/id/draft-kampanakis-curdle-ssh-pq-ke-05.html
+sealed class MLKem768X25519Sha256KeyExchange : Curve25519Sha256KeyExchange
```
tmds, line 19: Let's also remove the hash name from the class name, and apply the same to `SNtruPrime761X25519Sha512KeyExchange`.

The rationale for not including the hash, is that if we'd add variations based on the hash, we'd add an argument to this class to pass the different hash algorithms.

tmds:
Looks great @scott-xu!

tmds:
@scott-xu many thanks for contributing this and the other key exchange methods!!

PR #312, opened 2025-01-04
Add env var check for Windows ssh-agent pipe name

jborean93:
Adds logic to use the SSH_AUTH_SOCK environment variable when selecting the SSH Agent named pipe on Windows. If the env var is unset then the default openssh-ssh-agent is used instead.

It turns out that `ssh.exe` or other tools like `ssh-add.exe` on Windows will use the value of `SSH_AUTH_SOCK` env var so this change replicates that behaviour. It also adds some extra checks on the value to ensure it is in the form or `\\.\pipe\...`.

### DIFF
--- a/src/Tmds.Ssh/SshAgent.cs
+++ b/src/Tmds.Ssh/SshAgent.cs
@@ -33,9 +33,11 @@ namespace Tmds.Ssh
         {
             get
             {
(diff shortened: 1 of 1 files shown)


tmds:
This is part of the 0.11.0 release which was just uploaded to nuget.org.

PR #376, opened 2025-03-04
ssh example: add flag for allocating terminal.

tmds:


tmds:
@jborean93 I wonder if you'd be interesting in giving a shot at making this example work on Windows?

I think it involves using https://learn.microsoft.com/en-us/windows/console/setconsolemode with one or more of these flags: ENABLE_VIRTUAL_TERMINAL_INPUT, DISABLE_NEWLINE_AUTO_RETURN, ENABLE_PROCESSED_OUTPUT in order to configure the Windows console to use virtual terminal sequences (see https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences). You can find some code using those APIs here: https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine.Rendering/VirtualTerminalMode.cs.

When implemented, executing the example application with arguments `-t <server> nano` should give you a working, interactive full-screen editor.

jborean93:
I can certainly have a look but I probably don't have too much time in the next few weeks sorry.

tmds:
@jborean93 whenever you have some time, that would be great!

I've just added some Windows specific code to this PR in an attempt to implement a cancellable stdin reader. I wouldn't be surprised if it still has some bugs.

PR #403, opened 2025-06-22
Use BCL MLKem

scott-xu:
.NET 10 adds ML-Kem to BCL. For lower targets, it adds support from [Microsoft.Bcl.Cryptography](https://www.nuget.org/packages/Microsoft.Bcl.Cryptography/) nuget package.

See https://github.com/dotnet/runtime/issues/113508

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -2,6 +2,7 @@
   <ItemGroup>
     <!-- Tmds.Ssh dependencies -->
     <PackageVersion Include="BouncyCastle.Cryptography" Version="2.5.1" />
+    <PackageVersion Include="Microsoft.Bcl.Cryptography" Version="10.0.0-preview.5.25277.114" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4">
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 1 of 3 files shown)


tmds:
@scott-xu the PR is still in "draft", is this ready, or is there some change you'd like to make?

tmds:
Probably the package is going to be in preview until .NET 10 gets released.

Until then, perhaps we can do:
```
#if NET10
bool UseBclMLKem => MLKem.IsSupported;
#else
bool UseBclMLKem => false;
#endif
```

Then we'll use the in-box .NET 10 MLKem for the net10.0 target.

tmds:
I think the `net10.0` will typeforward to the shared framework. Anyway, I'm ok with merging this PR if we make it so that, for now, only `net10.0` uses the BCL MLKem.

tmds:
Can you make the proposed change from https://github.com/tmds/Tmds.Ssh/pull/403#issuecomment-3028621525 so we only use the BCL MLKem for `net10.0`?

And when .NET 10 is released, we'll undo that change and use it for the other target frameworks as well.

scott-xu:
@tmds There's only one `MLKemImplementation' which is named as `NotSupported` in [Microsoft.Bcl.Cryptography](https://www.nuget.org/packages/Microsoft.Bcl.Cryptography/) preview5.
image

scott-xu:
Cheers!

PR #471, opened 2026-05-12
fix(sftp): truncate overwritten remote file when upload source is smaller

ameyarajwade7:
Fixes [SftpClient.UploadFileAsync(..., overwrite: true)](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/8b640eef5a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) behavior when replacing an existing remote file with smaller source content.

Previously, overwrite wrote from offset 0 but could leave stale trailing bytes from the previous remote file if the new source was shorter.

### DIFF
--- a/src/Tmds.Ssh/SftpChannel.cs
+++ b/src/Tmds.Ssh/SftpChannel.cs
@@ -958,12 +958,16 @@ sealed partial class SftpChannel : IDisposable
             throw new SftpException(SftpError.NoSuchFile);
         }
 
+        ValueTask<long> initialLengthTask = overwrite ? remoteFile.GetLengthAsync(cancellationToken) : ValueTask.FromResult(0L);
+
         // Pipeline the writes when the source is a sync, seekable Stream.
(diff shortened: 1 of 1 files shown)


src/Tmds.Ssh/SftpChannel.cs
tmds: We can truncate immediately when we open an existing file:

```suggestion
        using SftpFile? remoteFile = (await OpenFileCoreAsync(workingDirectory, remotePath, (overwrite ? SftpOpenFlags.OpenOrCreate | SftpOpenFlags.Truncate : SftpOpenFlags.CreateNew) | SftpOpenFlags.Write, permissions, SftpClient.DefaultFileOpenOptions, cancellationToken).ConfigureAwait(false));
```

ameyarajwade7: Yes. we can do that. I have updated the pr.

tmds:
@ameyarajwade7, thanks!

PR #497, opened 2026-07-29
Expose SSH user authentication banners

buchmiet:
`SSH_MSG_USERAUTH_BANNER` packets are currently parsed and discarded. This leaves
applications unable to show server messages sent during authentication. Besides login
notices, some servers use the banner for an action the user must take: Tailscale SSH
check mode sends its browser sign-in URL in a banner and waits for that check to
complete.

This change adds:

- a `BannerMessageHandler` delegate and a `BannerMessageContext` carrying the message
  and `SshConnectionInfo`;
- `BannerMessageHandler` properties on `SshClientSettings` and `SshConfigSettings`,
  propagated through SSH config loading and proxy-jump settings;
- control character escaping in the library: TAB, CR, and LF are kept, other control
  characters are replaced by an octal escape sequence per UTF-8 byte, matching
  OpenSSH's banner output (RFC 4252 §5.4, RFC 4251 §9.2);
- strict payload parsing (the language tag is read and skipped) with guaranteed packet
  disposal.

The handler is synchronous and is invoked before the next authentication packet is
read. When no handler is configured, banners remain ignored, and `MaxBannerPackets`
remains enforced before parsing or invoking user code.

The default test server now sends a banner, so the existing suite also covers
connecting without a handler. A single integration test verifies delivery through the
delegate; unit tests cover parsing and escaping. The guide documents the API, and the
`ssh` and `ssh-cp` example apps write banners to stderr.

The full container-backed test suite passes on net8.0 through net11.0.

### DIFF
--- a/src/Tmds.Ssh/Banner.cs
+++ b/src/Tmds.Ssh/Banner.cs
@@ -6,12 +6,11 @@ namespace Tmds.Ssh;
 /// <summary>
 /// Context for a banner message sent by the server during authentication.
 /// </summary>
-public struct BannerContext
+public struct BannerMessageContext
 {
-    internal BannerContext(string message, string languageTag, SshConnectionInfo connectionInfo)
+    internal BannerMessageContext(string message, SshConnectionInfo connectionInfo)
     {
         Message = message;
-        LanguageTag = languageTag;
         ConnectionInfo = connectionInfo;
     }
 
@@ -19,40 +18,25 @@ public struct BannerContext
     /// Gets the banner message.
     /// </summary>
     /// <remarks>
-    /// The message is under the control of the server. It may be empty, span multiple lines, and
-    /// contain control characters. Treat it as untrusted when displaying it to a user.
+    /// The message may be empty and may span multiple lines. Control characters other than tab,
+    /// carriage return, and newline are replaced by escape sequences, so the message is safe to
+    /// write to a terminal.
     /// </remarks>
     public string Message { get; }
-
-    /// <summary>
-    /// Gets the language tag of the message, as described in RFC 3066. Often empty.
-    /// </summary>
-    public string LanguageTag { get; }
 
     /// <summary>
     /// Gets the SSH connection information.
     /// </summary>
     public SshConnectionInfo ConnectionInfo { get; }
-
-    /// <summary>
-    /// Returns whether batch (non-interactive) mode is enabled.
-    /// </summary>
-    /// <remarks>
-    /// In batch mode the <see cref="BannerHandler"/> delegate mustn't make interactive prompts.
-    /// </remarks>
-    public bool IsBatchMode => ConnectionInfo.IsBatchMode;
 }
 
 /// <summary>
 /// Delegate for handling banner messages sent by the server.
 /// </summary>
-/// <param name="context">The banner context.</param>
-/// <param name="cancellationToken">Token to cancel the operation.</param>
+/// <param name="context">The banner message context.</param>
 /// <remarks>
 /// The server may send a banner at any time after the authentication protocol starts and before
-/// authentication succeeds, and may send several. Authentication does not continue until the
-/// delegate completes, so a delegate that waits for the user also delays the connection. Bear in
-/// mind that <see cref="SshClientSettings.ConnectTimeout"/> bounds the entire authenticated
-/// connect, including time spent in this delegate.
+/// authentication succeeds, and may send several. The next packet is not read until the delegate
+/// returns.
 /// </remarks>
-public delegate ValueTask BannerHandler(BannerContext context, CancellationToken cancellationToken);
+public delegate void BannerMessageHandler(BannerMessageContext context);
--- a/src/Tmds.Ssh/Banner.cs
+++ b/src/Tmds.Ssh/Banner.cs
@@ -39,4 +39,4 @@ public struct BannerMessageContext
 /// authentication succeeds, and may send several. The next packet is not read until the delegate
 /// returns.
 /// </remarks>
-public delegate void BannerMessageHandler(BannerMessageContext context);
+public delegate void BannerHandler(BannerMessageContext context);
--- a/src/Tmds.Ssh/Banner.cs
+++ b/src/Tmds.Ssh/Banner.cs
@@ -1,0 +1,58 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
+namespace Tmds.Ssh;
+
+/// <summary>
+/// Context for a banner message sent by the server during authentication.
+/// </summary>
+public struct BannerContext
+{
+    internal BannerContext(string message, string languageTag, SshConnectionInfo connectionInfo)
+    {
+        Message = message;
+        LanguageTag = languageTag;
+        ConnectionInfo = connectionInfo;
+    }
+
+    /// <summary>
+    /// Gets the banner message.
+    /// </summary>
+    /// <remarks>
+    /// The message is under the control of the server. It may be empty, span multiple lines, and
+    /// contain control characters. Treat it as untrusted when displaying it to a user.
+    /// </remarks>
+    public string Message { get; }
+
+    /// <summary>
+    /// Gets the language tag of the message, as described in RFC 3066. Often empty.
+    /// </summary>
+    public string LanguageTag { get; }
+
+    /// <summary>
+    /// Gets the SSH connection information.
+    /// </summary>
+    public SshConnectionInfo ConnectionInfo { get; }
+
+    /// <summary>
+    /// Returns whether batch (non-interactive) mode is enabled.
+    /// </summary>
+    /// <remarks>
+    /// In batch mode the <see cref="BannerHandler"/> delegate mustn't make interactive prompts.
+    /// </remarks>
+    public bool IsBatchMode => ConnectionInfo.IsBatchMode;
+}
+
+/// <summary>
+/// Delegate for handling banner messages sent by the server.
+/// </summary>
+/// <param name="context">The banner context.</param>
+/// <param name="cancellationToken">Token to cancel the operation.</param>
+/// <remarks>
+/// The server may send a banner at any time after the authentication protocol starts and before
+/// authentication succeeds, and may send several. Authentication does not continue until the
+/// delegate completes, so a delegate that waits for the user also delays the connection. Bear in
+/// mind that <see cref="SshClientSettings.ConnectTimeout"/> bounds the entire authenticated
+/// connect, including time spent in this delegate.
+/// </remarks>
+public delegate ValueTask BannerHandler(BannerContext context, CancellationToken cancellationToken);
--- a/src/Tmds.Ssh/UserAuthContext.cs
+++ b/src/Tmds.Ssh/UserAuthContext.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Extensions.Logging;
 using System.Diagnostics;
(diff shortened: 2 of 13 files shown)


src/Tmds.Ssh/Banner.cs
```diff
@@ -0,0 +1,58 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
+namespace Tmds.Ssh;
+
+/// <summary>
+/// Context for a banner message sent by the server during authentication.
+/// </summary>
+public struct BannerContext
```
tmds: Let's name this `BannerMessageContext`.

src/Tmds.Ssh/Banner.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// </summary>
+    public SshConnectionInfo ConnectionInfo { get; }
+
+    /// <summary>
+    /// Returns whether batch (non-interactive) mode is enabled.
+    /// </summary>
+    /// <remarks>
+    /// In batch mode the <see cref="BannerHandler"/> delegate mustn't make interactive prompts.
+    /// </remarks>
+    public bool IsBatchMode => ConnectionInfo.IsBatchMode;
```
tmds: I'm not sure why this is here. Let's leave it out.

src/Tmds.Ssh/Banner.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// <remarks>
+    /// The message is under the control of the server. It may be empty, span multiple lines, and
+    /// contain control characters. Treat it as untrusted when displaying it to a user.
+    /// </remarks>
+    public string Message { get; }
+
+    /// <summary>
+    /// Gets the language tag of the message, as described in RFC 3066. Often empty.
+    /// </summary>
+    public string LanguageTag { get; }
```
tmds: Let's leave out the language tag. It is not used much in practice afaik and it can be added later.

src/Tmds.Ssh/Banner.cs
```diff
@@ -0,0 +1,58 @@
...
+/// <param name="context">The banner context.</param>
+/// <param name="cancellationToken">Token to cancel the operation.</param>
+/// <remarks>
+/// The server may send a banner at any time after the authentication protocol starts and before
+/// authentication succeeds, and may send several. Authentication does not continue until the
+/// delegate completes, so a delegate that waits for the user also delays the connection. Bear in
+/// mind that <see cref="SshClientSettings.ConnectTimeout"/> bounds the entire authenticated
+/// connect, including time spent in this delegate.
+/// </remarks>
+public delegate ValueTask BannerHandler(BannerContext context, CancellationToken cancellationToken);
```
tmds: Let's name this `BannerMessageHandler`.
Also let's change the signature to be `void` return and omit the `CancellationToken`.

src/Tmds.Ssh/Banner.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /// <summary>
+    /// Gets the banner message.
+    /// </summary>
+    /// <remarks>
+    /// The message is under the control of the server. It may be empty, span multiple lines, and
+    /// contain control characters. Treat it as untrusted when displaying it to a user.
+    /// </remarks>
+    public string Message { get; }
```
tmds, line 25: Let make the library responsible for handling control characters by escaping them.

test/Tmds.Ssh.Tests/BannerTests.cs
```diff
@@ -0,0 +1,133 @@
+using Xunit;
+
+namespace Tmds.Ssh.Tests;
+
+[Collection(nameof(BannerSshServerCollection))]
```
tmds: Server collections are expensive because they require their own container image.
Let's add a banner to the default server and have a single test that verifies it is returned through the delegate.

tmds:
@buchmiet thanks for picking this up! I've added a few comments. ptal.

Can you also include the API in the guide (docfx/index.md)? And use it in the `ssh` and `ssh-cp` example apps?

src/Tmds.Ssh/Banner.cs
buchmiet: Renamed to `BannerMessageContext`.

src/Tmds.Ssh/Banner.cs
buchmiet: Done. The library escapes control characters before invoking the handler, following OpenSSH's `fmprintf` semantics: TAB, CR, and LF are kept; other control characters (C0, DEL, C1) are replaced by an octal escape per UTF-8 byte, so e.g. `ESC[2J` is delivered as `\033[2J` (RFC 4252 §5.4, RFC 4251 §9.2). Unit tests cover NUL, BEL, ESC, DEL, and a C1 character.

src/Tmds.Ssh/Banner.cs
buchmiet: Removed from the public API. The parser still reads and skips the field so the packet remains fully validated.

src/Tmds.Ssh/Banner.cs
buchmiet: Removed. Agreed — the banner is a notification, not a prompt, so batch mode doesn't apply to it.

src/Tmds.Ssh/Banner.cs
buchmiet: Renamed to `BannerMessageHandler` and changed to `void` without a `CancellationToken`. I also renamed the settings properties to match, following the `HostAuthentication`/`PasswordPrompt` convention.

test/Tmds.Ssh.Tests/BannerTests.cs
buchmiet: Done. The default server image now sends a banner, the dedicated collection and the extra-mount support are removed, and a single test verifies delivery through the delegate. The rest of the suite now also implicitly covers connecting without a handler.

buchmiet:
Thanks for the review! All comments are addressed:

- `BannerMessageContext` / `BannerMessageHandler`, synchronous `void` delegate without
  a `CancellationToken`. I also renamed the settings properties to match the delegate
  name, following the `HostAuthentication`/`PasswordPrompt` convention.
- `LanguageTag` and `IsBatchMode` are gone. The language tag is still read so the
  packet stays fully validated.
- The library now escapes control characters before invoking the handler, following
  OpenSSH: TAB, CR, and LF are kept; other control characters are replaced by an octal
  escape sequence per UTF-8 byte (RFC 4251 §9.2).
- The dedicated server collection is removed: the default server now sends a banner
  and a single test verifies it is returned through the delegate.
- The guide has an "Authentication Banners" section, and `ssh`/`ssh-cp` write banners
  to stderr.

The full container-backed suite passes on net8.0 through net11.0. ptal!

src/docfx/index.md
```diff
@@ -304,6 +304,18 @@ Credentials = [ new KerberosCredential() ]
 Credentials = [ new KerberosCredential(new NetworkCredential("user", "password", "REALM"), delegateCredential: true) ]
 ```
 
+#### Authentication Banners
```
tmds: I don't think we need a whole section. Let's include it in the connection settings table.

You can order the rows alphabetically.

src/ssh/Program.cs
```diff
@@ -270,6 +270,9 @@ static SshConfigSettings CreateSshConfigSettings(string[] options)
     }
     configSettings.Options = optionsDict;
 
+    configSettings.BannerMessageHandler = (BannerMessageContext ctx) =>
```
tmds: Looking at the example, I want to trim down the name to be `BannerHandler` instead of `BannerMessageHandler` both for the delegate and the property.

I prefer to keep `BannerMessageContext` because I find it indicates that there may be multiple, rather than just one banner that is delivered as a whole.

src/Tmds.Ssh/UserAuthContext.cs
```diff
@@ -207,6 +231,62 @@ public void SetFailed()
...
+        string message = reader.ReadUtf8String();
+        reader.SkipString(); // language tag
+        reader.ReadEnd();
+
+        return new BannerMessageContext(EscapeControlCharacters(message), connectionInfo);
+    }
+
+    // Match OpenSSH: keep tab, carriage return, and newline; replace other control
+    // characters by an octal escape sequence per UTF-8 byte (RFC 4251 section 9.2).
+    internal static string EscapeControlCharacters(string message)
```
tmds: Can you base the implementation on https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleControlCharacterSanitizer.cs for detecting the characters to escape and the escape format?

src/Tmds.Ssh/UserAuthContext.cs
```diff
@@ -104,13 +111,30 @@ public async ValueTask<Packet> ReceivePacketAsync(CancellationToken ct, int maxL
                    time after this authentication protocol starts and before
                    authentication is successful. */
 
-                // TODO: provide the banner to the user.
-
-                packet.Dispose();
+                BannerMessageHandler? bannerMessageHandler = _bannerMessageHandler;
```
tmds: Let's log the banner to the `_logger` at `Information` level.

When the logger isn't enabled at that level, we want to avoid the overhead of parsing the banner.

tmds:
Thanks for addressing the feedback. I have added some additional comments. The implementation should be ready to merge once those are handled.

src/Tmds.Ssh/UserAuthContext.cs
tmds: You can ignore this request.

src/ssh/Program.cs
buchmiet: Renamed the delegate and both settings properties to `BannerHandler`. `BannerMessageContext` keeps its name.

src/Tmds.Ssh/UserAuthContext.cs
buchmiet: Done. The escaping now follows `ConsoleControlCharacterSanitizer`: the same characters are detected (C0, DEL, and C1, with tab, newline, and carriage return preserved) using `SearchValues`, and escape sequences are written as `\uXXXX`. It uses the `ValueStringBuilder` that is already in the repo. The escaping tests are updated for the new format.

src/docfx/index.md
buchmiet: Done. The section is replaced by a `BannerHandler` row in the connection settings table, and the table rows are now ordered alphabetically.

buchmiet:
Thanks! Addressed:

- The delegate and the `SshClientSettings`/`SshConfigSettings` properties are now
  `BannerHandler`; `BannerMessageContext` keeps its name.
- The escaping is based on `ConsoleControlCharacterSanitizer`: the same detection range,
  the `\uXXXX` escape format, and `SearchValues` for the scan.
- The guide section is replaced by a row in the connection settings table, which is now
  ordered alphabetically.

I left out the banner logging per your follow-up.

The full container-backed suite passes on net8.0 through net11.0. ptal!

tmds:
Thanks @buchmiet!

I'll post a message here when this is part of a release that landed on nuget.org.

Until then, you can use a CI package with version `0.23.1-*` from the CI feed: https://www.myget.org/F/tmds/api/v3/index.json,

buchmiet:
Thanks for the reviews and for merging this! I learned a lot.

tmds:
This is available as part of the 0.24.0 release which was just uploaded to nuget.org.

PR #510, opened 2026-09-09
Support SSH agent forwarding

### LINKED ISSUE #511: full support SSH agent forwarding
full support SSH agent forwarding #510

joesdu:
Support SSH agent forwarding.
Adds the client side of 'ssh -A': the library requests agent forwarding on
session channels using 'auth-agent-req@openssh.com' and handles the
'auth-agent@openssh.com' channels the server opens in response.

- SshClientSettings.ForwardAgent enables forwarding. The agent channels the
  server opens are proxied to the local SSH agent.
- SshClientSettings.AgentChannelHandler hands those channels to the caller as
  an SshDataStream instead. This allows applying a policy to the requests made
  by the server, like only allowing the server to sign.
- Agent connections are bound to the session using 'session-bind@openssh.com'.
  This enables the agent to apply the constraints of keys that are restricted
  to specific destinations ('ssh-add -h'). It is also done for the connections
  that are used for authentication.
- The 'ForwardAgent' ssh config option is now understood. Besides yes/no, the
  forms that specify the agent socket path are supported.
- The 'ssh' tool accepts '-A'.

Close #511


tmds:
hi @joesdu,  I am quite loaded with work so it may take some time until I review this. I'm going to add some comments about the public API already.

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -432,6 +432,36 @@ public List<string> GlobalKnownHostsFilePaths
...
+
+    /// <summary>
+    /// Gets or sets the <see cref="Tmds.Ssh.AgentChannelHandler"/> delegate.
+    /// </summary>
+    /// <remarks>
+    /// <para>When set, the delegate handles the agent channels that are opened by the server while
+    /// <see cref="ForwardAgent"/> is enabled. When unset, those channels are proxied to the local SSH agent.</para>
+    /// <para>The delegate enables applying a policy to the requests made by the server, for example, to only allow signing.</para>
+    /// </remarks>
+    public AgentChannelHandler? AgentChannelHandler { get; set; }
```
tmds: I suggest leaving this out. Applying additional policies by handling the agent level protocol is really an advanced user feature. The session binding should provide the destination constraints.

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -432,6 +432,36 @@ public List<string> GlobalKnownHostsFilePaths
...
+
+    // Address of the agent to forward. When null, SshAgent.DefaultAddress is used.
+    // This is set when the 'ForwardAgent' config option specifies a path.
+    internal string? ForwardAgentAddress { get; set; }
```
tmds: Let's make this one public, and treat null and empty string as using the default SSH_AUTH_SOCK.

src/Tmds.Ssh/AgentForwarding.cs
```diff
@@ -0,0 +1,53 @@
...
+    private const int BufferSize = 4096;
+
+    // Proxies an agent channel opened by the server to the SSH agent at 'address'.
+    public static async Task ProxyToLocalAgentAsync(SshDataStream channel, string address, SshConnectionInfo connectionInfo, CancellationToken cancellationToken)
+    {
+        using Stream agentStream = await SshAgent.OpenStreamAsync(address, cancellationToken).ConfigureAwait(false);
+
+        // Binding tells the agent this connection is forwarded, which enables it to
+        // apply the constraints of keys that are restricted to specific destinations.
+        await SshAgent.TryBindSessionAsync(agentStream, connectionInfo, isForwarding: true, cancellationToken).ConfigureAwait(false);
```
tmds: We should log when we fail to bind.

src/Tmds.Ssh/SshClientSettings.cs
joesdu: Makes sense, I'll drop it. Forwarded channels will always be proxied to the local agent, and I'll remove the `AgentChannelHandler` delegate type, the docs section and the test that covered it. If filtering the agent protocol turns out to be needed, it can be added later without breaking anything.

I'll keep the warning in the `ForwardAgent` docs that forwarding gives anyone with access to the agent socket on the server control over the local agent, since session binding only constrains keys that were added with destination constraints (`ssh-add -h`).

src/Tmds.Ssh/SshClientSettings.cs
joesdu: Will do — `public string? ForwardAgentAddress { get; set; }`, with both `null` and `""` meaning the default agent (`SSH_AUTH_SOCK`, or the `openssh-ssh-agent` named pipe on Windows). I'll update the public API baseline and the settings table in the docs too.

src/Tmds.Ssh/AgentForwarding.cs
joesdu: I'll pass the logger into `SshAgent.TryBindSessionAsync` and log there, so it covers both the forwarded connections and the ones used for authentication.

I'm planning `LogLevel.Information` rather than `Warning`: agents that don't implement `session-bind@openssh.com` simply answer `SSH_AGENT_FAILURE`, so it isn't necessarily a problem (OpenSSH logs this at debug level). Happy to make it a warning if you'd rather have it stand out.

src/docfx/index.md
```diff
@@ -603,6 +605,38 @@ Pass port `0` to let the server assign a port. The actual assigned port is avail
 using var listener = await sshClient.ListenUnixAsync("/tmp/my.sock");
 ```
 
+
+### Forwarding the SSH agent
```
tmds: Let's drop this section. Mentioning the ForwardAgent and ForwardAgentAddress is sufficient. Many other sources are available to explain these further and there are no special considerations that apply only to the library.

src/docfx/index.md
```diff
@@ -343,6 +343,8 @@ When using <xref:Tmds.Ssh.SshConfigSettings>, the <xref:Tmds.Ssh.SshConfigSettin
 | `DefaultWindowSize` | 2 MB | SSH channel window size; larger values improve throughput on high-latency or high-bandwidth links at the cost of memory. It can be overridden per operation using `SftpClientOptions`, `ExecuteOptions`, or a `windowSize` method argument. |
 | `EnableBatchModeWhenConsoleIsRedirected` | `true` | Automatically enable batch mode when the console is redirected. |
 | `EnvironmentVariables` | | Environment variables set for all remote processes. |
+| `ForwardAgent` | `false` | Forward the local SSH agent to the server. See [Forwarding the SSH agent](#forwarding-the-ssh-agent). |
+| `ForwardAgentAddress` | | Address of the SSH agent to forward. When unset, the default SSH agent is forwarded. |
```
tmds: nit: let's say 'is used' instead of 'is forwarded'.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -34,7 +37,7 @@ public static string? DefaultAddress
             get
             {
                 string? defaultAddress = OperatingSystem.IsWindows()
-                    ? @"\\pipe\.\openssh-ssh-agent"
+                    ? @"\\.\pipe\openssh-ssh-agent"
```
tmds, line 40: Thanks for fixing!

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -43,25 +46,52 @@ public static string? DefaultAddress
...
+        public SshAgent(string address, SequencePool sequencePool, ILogger<SshClient> logger)
         {
             _address = address;
             _sequencePool = sequencePool;
+            _logger = logger;
         }
 
-        public async ValueTask ConnectAsync(CancellationToken cancellationToken)
+        // When 'bindSession' is set, the connection is bound to the session (see TryBindSessionAsync).
+        public async ValueTask ConnectAsync(SshConnectionInfo? bindSession, CancellationToken cancellationToken)
```
tmds: Let's make the SshConnectionInfo a non nullable argument of the SshAgent together with a bool that says isForwarding.

Then we can make private the TryBindSessionAsync and OpenStreamAsync.

src/Tmds.Ssh/SshConfigParser.cs
```diff
@@ -514,9 +516,34 @@ we have some basic handling that checks the option value indicates the feature i
...
+                    }
+                    else if (value.Equals("yes", StringComparison.OrdinalIgnoreCase))
+                    {
+                        config.ForwardAgent = true;
+                    }
+                    else
+                    {
+                        // The value may also be the path of the agent socket, or the name of an
+                        // environment variable (prefixed with '$') that holds that path.
+                        string? address = value.StartsWith("$")
```
tmds: envvar expansion isn't supported overall. Let's leave it out here with a TODO in comments.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -678,6 +694,35 @@ uint32    originator port
         TrySendPacket(_sequencePool.CreateChannelOpenFailureMessage(remoteChannel));
     }
 
+    // Agent forwarding is enabled when the user asked for it and we know which agent to forward.
+    private static bool IsAgentForwardingEnabled([NotNullWhen(true)] SshClientSettings? settings)
```
tmds: Let's add an arg `string? _forwardAgentAddress` that we set once and keep at `null` when forwarding is not enabled.

src/Tmds.Ssh/AgentForwarding.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Tmds.Ssh;
+
+// Implements the client side of 'auth-agent-req@openssh.com'/'auth-agent@openssh.com' agent forwarding.
+static class AgentForwarding
+{
+    private const int BufferSize = 4096;
+
+    // Proxies an agent channel opened by the server to the SSH agent at 'address'.
+    public static async Task ProxyToLocalAgentAsync(SshDataStream channel, string address, SshConnectionInfo connectionInfo, ILogger<SshClient> logger, CancellationToken cancellationToken)
```
tmds: Let's move this logic in the `SshSession` class and change the implementation for forwarding streams to be close to the existing `ForwardConnectionAsync` (or reuse it if possible).

src/Tmds.Ssh/KeyExchange.cs
```diff
@@ -44,6 +44,13 @@ public async Task<KeyExchangeOutput> TryExchangeAsync(KeyExchangeContext context
 
             VerifySignature(connectionInfo.ServerKey, input.HostKeyAlgorithms, exchangeHash, serverReply.exchangeHashSignature, connectionInfo);
 
+            if (connectionInfo.SessionId is null)
```
tmds: Rather than make this depend on `SessionId is null`, it would be clearer to assign using `??=`.

src/Tmds.Ssh/SshAgent.cs
tmds: Let's refactor this a bit:
- drop the logger from the constructor
- drop the SshConnectionInfo from ConnectAsync and remove OpenStreamAsync (since that is what ConnectAsync does)
- Change TryBindSessionAsync to be a public instance method that accepts the SshConnectionInfo and a bool isForward argument. Its callers are responsible for using the logger.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -638,6 +638,22 @@ string          socket path
...
+            // The server opens these channels in response to 'auth-agent-req@openssh.com'.
+            SshClientSettings? settings = _settings;
+            if (IsAgentForwardingEnabled(settings))
+            {
+                SshChannel channel = CreateChannel(typeof(SshDataStream), windowSize: null, onAbort: null, remoteChannel, checked((int)maxPacketSize), checked((int)initialWindowSize));
+                channel.TrySendChannelOpenConfirmationMessage(remoteChannel);
+
+                // Don't handle the channel on the receive loop.
+                SshDataStream stream = new SshDataStream(channel);
+                _ = Task.Run(() => HandleAgentChannelAsync(stream, settings));
```
tmds: I don't think we need a Task.Run here?

src/Tmds.Ssh/SshSession.cs
```diff
@@ -906,6 +951,13 @@ public async Task OpenSessionAsync(SshChannel channel, ExecuteOptions? options,
             await channel.ReceiveChannelOpenConfirmationAsync(cancellationToken).ConfigureAwait(false);
         }
 
+        if (IsAgentForwardingEnabled(_settings))
+        {
+            // Like 'ssh' we don't wait for a reply.
```
tmds, line 1070: We may want to check here if there is actually an agent we can connect to before announcing it to the server?

src/Tmds.Ssh/SshSession.cs
```diff
@@ -638,6 +638,22 @@ string          socket path
             string path = reader.ReadUtf8String();
             listenAddress = new ListenAddress(channelType, path, 0);
         }
+        else if (channelType == AlgorithmNames.AuthAgent)
+        {
+            // The server opens these channels in response to 'auth-agent-req@openssh.com'.
+            SshClientSettings? settings = _settings;
+            if (IsAgentForwardingEnabled(settings))
```
tmds: If we can't connect to the agent, let's return CHANNEL_OPEN_FAILURE.

src/Tmds.Ssh/SshAgent.cs
```diff
@@ -44,6 +47,7 @@ public static string? DefaultAddress
         private readonly string _address;
         private readonly SequencePool _sequencePool;
 
+        private Stream? _stream;
```
tmds: Let's pass the Stream as a constructor argument, wrap it into a StreamSshConnection, and assign it to _agentConnection. That removes _stream, the public Stream property, and GetAgentConnection().

Let's be explicit about the ownership by using an ownsStream arg too in the constructor.

.gitignore
```diff
@@ -366,4 +366,9 @@ FodyWeavers.xsd
...
 # DocFX
 _site/
-src/docfx/api/
\ No newline at end of file
+src/docfx/api/
+
+# AI Agents
+.omo
+.omc
+.sisyphus
```
tmds: nit: please leave these out.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -678,6 +700,142 @@ uint32    originator port
...
+        try
+        {
+            using SshAgent agent = new SshAgent(_forwardAgentAddress, _sequencePool);
+            await agent.ConnectAsync(ct).ConfigureAwait(false);
+
+            return true;
+        }
+        catch (Exception ex) when (!ct.IsCancellationRequested)
+        {
+            Logger.CannotConnectToSshAgent(ex);
```
tmds: Let's do a different message here: SkippingAgentForwarding with Information level and a message that ways "Skipping agent forwarding, cannot connect to SSH agent".

src/Tmds.Ssh/SshSession.cs
```diff
@@ -906,6 +1064,14 @@ public async Task OpenSessionAsync(SshChannel channel, ExecuteOptions? options,
             await channel.ReceiveChannelOpenConfirmationAsync(cancellationToken).ConfigureAwait(false);
         }
 
+        // Only announce agent forwarding when there is an agent we can forward.
+        if (_forwardAgentAddress is not null && await CanConnectToForwardAgentAsync(cancellationToken).ConfigureAwait(false))
+        {
+            // Like 'ssh' we don't wait for a reply.
+            // When the server refuses, it will not open any agent channels.
+            channel.TrySendAuthAgentRequestMessage();
```
tmds, line 1058: Let's log at Information level here "Enabling agent forwarding"

src/Tmds.Ssh/SshSession.cs
```diff
@@ -678,6 +700,142 @@ uint32    originator port
...
+            {
+                Logger.CannotConnectToSshAgent(ex);
+            }
+            TrySendPacket(_sequencePool.CreateChannelOpenFailureMessage(remoteChannel));
+            return;
+        }
+
+        try
+        {
+            using (agent)
```
tmds, line 728: Let's log here at Information level: "Forwarding agent channel".

tmds:
@joesdu I've added a few more comments, we are nearing the finish line...

src/Tmds.Ssh/SshSession.cs
joesdu: 👌

src/Tmds.Ssh/SshSession.cs
joesdu: 👌

tmds:
I made some minor changes to the SshAgent. My feedback wasn't working well because I didn't fully understand the interactions.

I also rebased the PR due to a bug in the .NET 11 RC1 which caused CI issues (https://github.com/tmds/Tmds.Ssh/pull/517).

This is good to merge now. Thanks for contributing this feature @joesdu!

PR #513, opened 2026-09-10
Add zlib@openssh.com compression.

joesdu:
Implements the ~~`zlib` (RFC 4253) and~~ `zlib@openssh.com` compression algorithm~~s~~.

Compression is applied to the packet payload before it gets encrypted, by wrapping the `IPacketEncryptor`/`IPacketDecryptor` that are created during the key exchange.

## Behavior

- Like OpenSSH, compression is **not enabled by default**: `DefaultCompressionAlgorithms` stays `none`.
- `SshClientSettings.CompressionAlgorithmsClientToServer` and `CompressionAlgorithmsServerToClient` are now `public` (they were `internal`).
- `SshConfigOption.Compression` was added, so it can also be set through `SshConfigSettings.Options`. The `Compression` config file keyword was already parsed; it now maps onto the algorithms instead of being a no-op.
- `zlib@openssh.com` only starts compressing after `SSH_MSG_USERAUTH_SUCCESS` is received, for both directions. When rekeying after authentication it compresses from the start of the new keys.
- The compression context is reset on each key exchange, which matches OpenSSH.
- The decompressed payload is bounded by the maximum packet length so a peer can not force an arbitrary allocation.

## Implementation

`ZLibStream` is used for both directions, so deflate and inflate run on the runtime's native zlib. A single stream spans the lifetime of the keys; each payload is flushed so the peer can decompress it right away while the context is kept for the next payload. `Flush()` performs a sync flush where OpenSSH uses a partial flush, which costs 4 bytes per packet and is decompressed by either side.

The payload segments are written into the deflater directly and the inflated data is appended to the packet `Sequence` directly, so no intermediate copies are needed.

### `System.IO.Compression.UseStrictValidation`

The zlib stream is flushed per payload and never ends, so the read that finds no more decompressed data reports the stream as truncated when an application enables the [`System.IO.Compression.UseStrictValidation`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.IO.Compression/src/System/IO/Compression/DeflateZLib/DeflateStream.cs) switch, which is off by default.

That read returns no data of its own (the payload was returned by the preceding reads), so it is treated as the end of the payload. Data that decompresses to nothing is still reported as an error, so a payload that is not valid zlib data does not pass silently.

I ran the tests with the switch both off and on: all of them pass either way.

### `Sequence.Append`

`Sequence.AllocGetSpan(n)` starts a new segment when the current one has less than `n` bytes left. Appending decompressed data that way leaves the 5 header bytes alone in the first segment, while `Packet.MessageId` reads the message id from `FirstSpan`. `Sequence.Append(ReadOnlySpan<byte>)` fills up the current buffer before allocating a new one, which keeps the header and the message id contiguous.

## Tests

- `ZLibCompressionTests`: round-trips, incompressible payloads, compression context retained across payloads, the decompressed length bound, invalid data, and that a decompressed packet spanning several buffers has a readable message id.
- `CompressionTests`: negotiation (both directions, unknown names skipped, no common algorithm), data round-trips up to 100KB of compressible and incompressible data, and enabling compression through `SshConfigOption.Compression`.
- `RekeyTests.HandlesServerInitiatedRekey` is now a theory that also runs with compression enabled, which covers resetting the compression context on rekey.

The integration tests run against the sshd container, which offers `none,zlib@openssh.com`. ~~OpenSSH's server does not offer plain `zlib`, so that one is covered by the unit tests only.~~

### DIFF
--- a/src/Tmds.Ssh/SshClientSettings.cs
+++ b/src/Tmds.Ssh/SshClientSettings.cs
@@ -593,7 +593,11 @@ public sealed partial class SshClientSettings
     /// <summary>
     /// Gets or sets the permitted compression algorithms for client to server communication in order of preference.
     /// </summary>
-    internal AlgorithmList CompressionAlgorithmsClientToServer
+    /// <remarks>
+    /// By default only <c>none</c> is permitted, which means the packets are not compressed.
+    /// Compression is enabled by including <c>zlib@openssh.com</c> (which starts compressing after authentication) or <c>zlib</c>.
+    /// </remarks>
+    public AlgorithmList CompressionAlgorithmsClientToServer
     {
         get => _compressionAlgorithmsClientToServer ??= new AlgorithmList(DefaultCompressionAlgorithms);
         set => _compressionAlgorithmsClientToServer = value;
@@ -602,7 +606,11 @@ public sealed partial class SshClientSettings
     /// <summary>
     /// Gets or sets the permitted compression algorithms for server to client communication in order of preference.
     /// </summary>
-    internal AlgorithmList CompressionAlgorithmsServerToClient
+    /// <remarks>
+    /// By default only <c>none</c> is permitted, which means the packets are not compressed.
+    /// Compression is enabled by including <c>zlib@openssh.com</c> (which starts compressing after authentication) or <c>zlib</c>.
+    /// </remarks>
+    public AlgorithmList CompressionAlgorithmsServerToClient
     {
         get => _compressionAlgorithmsServerToClient ??= new AlgorithmList(DefaultCompressionAlgorithms);
         set => _compressionAlgorithmsServerToClient = value;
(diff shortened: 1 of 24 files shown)


tmds:
> zlib and zlib@openssh.com

Let's leave out zlib and only do the more secure variant.

joesdu:
> > zlib and [zlib@openssh.com](mailto:zlib@openssh.com)zlib 以及 zlib@openssh.com
> 
> Let's leave out zlib and only do the more secure variant.

OK,I will remove it.

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -593,7 +593,11 @@ public AlgorithmList MacAlgorithmsServerToClient
     /// <summary>
     /// Gets or sets the permitted compression algorithms for client to server communication in order of preference.
     /// </summary>
-    internal AlgorithmList CompressionAlgorithmsClientToServer
+    /// <remarks>
```
tmds, line 608: Thinking a bit about the API here, I'd like an easy way for users to enable compression without them having to think about compression algorithms.

That would be a `bool EnableCompression = false`.

The algorithm lists properties would default to `[ zlib@openssh.com, none ]`, but they would not be used unless `EnableCompression` gets set to `true`.

joesdu: 👌

Issue #512, opened 2026-09-10
Proposal: Add X11 forwarding support


Aecric:
Hello,

I would like to ask whether X11 forwarding would be considered for Tmds.Ssh.

I am working on a cross-platform terminal client, and would like to support Linux graphical application forwarding similar to:

```bash
ssh -X user@host
```

The current prototype implements:

- Sending the SSH `x11-req` channel request
- Accepting incoming `x11` channels
- MIT-MAGIC-COOKIE-1 authentication handling
- Forwarding connections to a local X11 Unix socket or TCP endpoint
- Session and channel lifecycle cleanup
- Connection failure handling and setup timeouts

The prototype is available in my fork:

https://github.com/Aecric/Tmds.Ssh/tree/feature/x11-forwarding

The implementation currently passes the existing test suite, additional X11 protocol tests, and OpenSSH Docker integration tests across .NET 8, 9, 10, and 11.

Before opening a pull request, I would like to understand the preferred API direction:

1. Should Tmds.Ssh provide a high-level X11 forwarding API, including the local X server connection?
2. Or should the library only expose the SSH X11 channel and authentication primitives, leaving platform-specific connection logic to the host application?
3. Are there any API, security, or compatibility requirements that should be considered before preparing a pull request?

I would appreciate your feedback on whether this feature fits the scope of Tmds.Ssh and which API design would be acceptable upstream. I am happy to adjust or split the implementation according to the project’s design preferences.

joesdu:
Link [#417](https://github.com/joesdu/VelaShell/issues/417)

PR #514, opened 2026-09-13
Add X11 forwarding

joesdu:
## Summary

This adds X11 forwarding, following the OpenSSH `ssh -X`/`-Y` behavior.

> Updated per https://github.com/tmds/Tmds.Ssh/pull/514#issuecomment-5675611238: the Wayland/waypipe part was removed from this PR and is proposed separately as an example rather than a library feature.

```csharp
var settings = new SshClientSettings("user@host")
{
    ForwardX11 = true,          // request X11 forwarding for remote processes
    ForwardX11Trusted = true,   // default: false (untrusted, uses 'xauth generate')
    // X11Display = ":0",       // default: DISPLAY
};
using var client = new SshClient(settings);
using var process = await client.ExecuteAsync("xterm", new ExecuteOptions { ForwardX11 = true });
```

- Sends an `x11-req` (after `pty-req`, before `env`) with a random fake `MIT-MAGIC-COOKIE-1` cookie for exec, shell and subsystem sessions. SFTP sessions don't inherit `SshClientSettings.ForwardX11`.
- Accepts server-opened `x11` channels only when X11 forwarding was requested on the connection. The X11 connection setup message is parsed (both byte orders), the fake cookie is verified (constant-time compare) and replaced by the real cookie, and the connection is forwarded to the local display. Connections with a non-matching cookie are rejected. Each display gets its own fake cookie, so multiple displays can be forwarded on one connection.
- Display parsing: `:0`, `:10.2`, `unix:0`, `host:N`, IPv6 hosts, and macOS launchd socket paths. Local displays connect to the Linux abstract socket, then `/tmp/.X11-unix/X<N>`; on Windows (e.g. VcXsrv) they connect to TCP `localhost:6000+N`.
- Authentication data:
  - trusted: read from the Xauthority file (`XAUTHORITY` or `~/.Xauthority`) without running an external program. When there is no entry, random data is used (like OpenSSH).
  - untrusted: generated with `xauth generate <display> MIT-MAGIC-COOKIE-1 untrusted timeout <n>`; connections are refused after `ForwardX11Timeout` (default 20 minutes). The xauth invocation is bounded by a 30 second timeout.
- ssh_config: `ForwardX11`, `ForwardX11Trusted`, `ForwardX11Timeout` (time format like `1h30m`) and `XAuthLocation` are now supported (`XAuthLocation` previously threw "Unsupported keyword").
- Failure handling: when `ExecuteOptions.ForwardX11` is explicitly `true`, setup failures throw (`SshOperationException` / `SshChannelException`). When X11 forwarding is only enabled through `SshClientSettings.ForwardX11` (e.g. `ForwardX11 yes` in `~/.ssh/config`), failures are logged and the process starts without X11 forwarding, so existing configurations don't start failing.

### Public API additions

- `ExecuteOptions`: `bool? ForwardX11`
- `SshClientSettings`: `ForwardX11`, `ForwardX11Trusted`, `ForwardX11Timeout`, `X11Display`, `XAuthLocation`
- `SshConfigOption`: `ForwardX11`, `ForwardX11Trusted`, `ForwardX11Timeout`, `XAuthLocation` (appended, existing values unchanged)

### Test infrastructure changes

- The test sshd image installs `xorg-x11-xauth` and enables `X11Forwarding yes`.
- `SshServer.PickFreePort` picks a port in 20000-40000 instead of the dynamic range: Docker Desktop on Windows fails to publish ports from the dynamic range (49152+), which made the fixture fail with "Connection refused". Happy to split this out if you prefer.

## Limitations

- Only `MIT-MAGIC-COOKIE-1` authentication is supported (`XDM-AUTHORIZATION-1` entries are ignored), like OpenSSH.
- Untrusted forwarding (the default) requires `xauth` on the client and an X server with the SECURITY extension. On systems without `xauth` (e.g. Windows), use `ForwardX11Trusted = true`.
- The server needs `X11Forwarding yes` and `xauth`.
- Trusted forwarding gives remote X clients full access to the local display.

## Test environment

- Windows 11 Pro 10.0.29667 (x64), .NET SDK 10.0.401
- Docker Desktop 29.7.2 (Linux containers); test sshd image: CentOS Stream 10, OpenSSH, xorg-x11-xauth
- WSL2 Debian 13 with WSLg (Weston compositor + XWayland), mirrored networking; tests run as a self-contained linux-x64 build

### Unit tests (no server)
- X11 display parsing (valid/invalid names, xauth display names).
- Xauthority parsing and matching (local hostname, IPv4, wildcard family, truncated files, missing file).
- X11 setup message authentication (both byte orders, unknown cookie, other protocol, invalid messages, expiry).
- ssh_config parsing for the new options and the time format, settings defaults.

### Integration tests (Docker sshd)
- Forwarding to a local fake X server with verification that the fake cookie is replaced by the real cookie; rejection of connections with an invalid cookie; `SshClientSettings.ForwardX11` / `ExecuteOptions.ForwardX11` combinations; explicit requests fail on setup errors while settings-only requests fall back; untrusted forwarding without xauth.

### Manual end-to-end
- `xterm` in the container forwarded to the WSLg X server (bridged to TCP), including two concurrent X11 connections; both closed cleanly with exit code 0.

---

Wayland forwarding through waypipe is proposed separately, as an addition to the `ssh` example application.


joesdu:
LINK: #512 
[VelaShell #417](https://github.com/joesdu/VelaShell/issues/417)

tmds:
> Add X11 forwarding and Wayland

For implementation we should treat these separate.

Let's focus this PR on adding X11.

For Wayland, it depends on a 3rd party dependency (`waypipe`) and I think a user can already do this himself using existing APIs. Perhaps we can extend the `ssh` example application (in a copy-pastable form) to demonstrate how to do this rather than make it a library feature.

joesdu:
Thanks for the feedback, that makes sense. I've split it up:

- This PR now only adds X11 forwarding. The waypipe code, the `ForwardWayland`/`Waypipe*` API and the Wayland tests are gone, and the test image no longer installs waypipe.
- #515 extends the `ssh` example application with a `--forward-wayland` option instead. It's implemented in `src/ssh/WaypipeForward.cs` using only the public API (`StartRemoteForwardAsync` + `RemoteUnixEndPoint`), so it can be copied into an application. No library or public API changes there.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -52,6 +53,11 @@ internal struct GlobalRequestReply
 
     public SshConnectionInfo ConnectionInfo { get; }
 
+    // Available once connected.
+    internal SshClientSettings Settings => _settings ?? throw new InvalidOperationException("Not connected.");
+
+    internal SshLoggers Loggers => _loggers;
```
tmds: Nothing seems to be using these. Let's remove them.

test/Tmds.Ssh.Tests/SshServer.cs
```diff
@@ -187,9 +187,22 @@ protected SshServer(string? sshdConfig, IMessageSink messageSink, string userPas
 
         static int PickFreePort(IPAddress interfaceAddress)
         {
-            using var s = new Socket(interfaceAddress.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
-            s.Bind(new IPEndPoint(interfaceAddress, 0));
-            return (s.LocalEndPoint as IPEndPoint)!.Port;
+            // Pick a port below the dynamic port range (49152+) because Docker Desktop on Windows
```
tmds: Can you keep the existing behavior on non-Windows?

joesdu: Okay, I'll make some adjustments.

src/Tmds.Ssh/SshSession.cs
joesdu: 👌

src/Tmds.Ssh/SshSession.cs
```diff
@@ -660,6 +666,30 @@ string          socket path
...
+        {
+            /*
+                string    originator address (e.g., "192.168.7.38")
+                uint32    originator port
+            */
+            string originatorAddress = reader.ReadUtf8String();
+            uint originatorPort = reader.ReadUInt32();
+
+            X11Forwarding? x11Forwarding;
+            lock (_gate)
```
tmds: This shouldn't need a lock because we can only have to handle an x11 channel if we've ran through the assignment prior.

joesdu: 👌

src/Tmds.Ssh/SshSession.cs
```diff
@@ -1038,7 +1061,7 @@ public async Task<ISshChannel> OpenRemoteShellChannelAsync(Type channelType, Exe
         }
     }
 
-    public async Task OpenSessionAsync(SshChannel channel, ExecuteOptions? options, CancellationToken cancellationToken)
+    public async Task OpenSessionAsync(SshChannel channel, ExecuteOptions? options, CancellationToken cancellationToken, bool inheritX11Settings = true)
```
tmds: Instead of `inheritX11Settings`, can we have a static ExecuteOptions that we use for sftp subsystem which disables the X11?

src/Tmds.Ssh/SshSession.cs
```diff
@@ -1066,6 +1089,13 @@ public async Task OpenSessionAsync(SshChannel channel, ExecuteOptions? options,
             await channel.ReceiveChannelRequestSuccessAsync("Failed to allocate pseudoterminal.", cancellationToken).ConfigureAwait(false);
         }
 
+        bool? forwardX11 = options?.ForwardX11;
+        if (forwardX11 ?? (inheritX11Settings && _settings.ForwardX11))
+        {
+            // When X11 forwarding is enabled through the settings, failures don't fail the operation.
+            await RequestX11ForwardingAsync(channel, isRequired: forwardX11 == true, cancellationToken).ConfigureAwait(false);
+        }
```
tmds, line 1097: I think we can drop the `inheritX11Settings` arg, and do something like:
```
bool shouldForward = options?.ForwardX11 ?? _settings.ForwardX11;
if (shouldForward)
{
    await SetupX11ForwardingAsync(channel, isRequired: options?.ForwardX11 == true, cancellationToken);
}
```

src/Tmds.Ssh/ExecuteOptions.cs
```diff
@@ -142,6 +142,17 @@ public Dictionary<string, string> EnvironmentVariables
...
+    /// <summary>
+    /// Gets or sets whether to request X11 forwarding.
+    /// </summary>
+    /// <remarks>
+    /// <para>When <see langword="null"/> (the default), <see cref="SshClientSettings.ForwardX11"/> is used.</para>
+    /// <para>When set to <see langword="true"/>, the operation fails when X11 forwarding can not be set up.
+    /// When X11 forwarding is enabled through <see cref="SshClientSettings.ForwardX11"/>, failures are logged and the remote process is started without X11 forwarding.</para>
+    /// <para>The forwarded display and authentication are configured using <see cref="SshClientSettings.X11Display"/>, <see cref="SshClientSettings.ForwardX11Trusted"/>, <see cref="SshClientSettings.ForwardX11Timeout"/>, and <see cref="SshClientSettings.XAuthLocation"/>.</para>
+    /// </remarks>
+    public bool? ForwardX11 { get; set; }
```
tmds: It would make sense to provide the same options for the agent forward. Out of scope for this PR.

src/Tmds.Ssh/SshSession.cs
joesdu: Already fixed in `cf6b845`

src/Tmds.Ssh/ExecuteOptions.cs
tmds: What are your thoughts on making this an enum, something like:

```cs
enum ForwardMode
{
    // Use SshClientSettings Forward setting: Request when set to true.
    Default,
    // Do not forward, regardless of settings.
    Off,
    // Forward. Log and continue when setup fails.
    Request,
    // Forward. Throw when setup fails.
    Require,
}
```

tmds: Let's change the API here to be (nullable):

```
public ForwardMode? ForwardX11 { get; set; }
```
and on `SshClientSettings` to be (non-nullable):
```
public ForwardMode ForwardX11 { get; set; }
```
where:
```cs
enum ForwardMode
{
    /// Do not forward.
    Off,
    /// Forward. Log and continue when setup fails.
    Request,
    /// Forward. Throw when setup fails.
    Require,
}
```
And the effective forward mode is: `executeOptions?.ForwardX11 ?? clientSettings.ForwardX11`.

joesdu: 👌

src/Tmds.Ssh/ForwardMode.cs
```diff
@@ -0,0 +1,25 @@
+// This file is part of Tmds.Ssh which is released under MIT.
+// See file LICENSE for full license details.
+
+namespace Tmds.Ssh;
+
+/// <summary>
+/// Specifies whether X11 forwarding is enabled and how setup failures are handled.
```
tmds: Let's drop "X11" in these descriptions.

src/Tmds.Ssh/ForwardMode.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    Off,
+
+    /// <summary>
+    /// Forward. Log and continue when setup fails.
+    /// </summary>
+    Request,
+
+    /// <summary>
+    /// Forward. Throw when setup fails.
```
tmds: Throw -> Fail

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -536,6 +538,72 @@ public int DefaultWindowSize
         }
     }
 
+    /// <summary>
+    /// Gets or sets whether to request X11 forwarding for remote processes.
```
tmds: "Gets or sets whether to forward X11 connections"

src/Tmds.Ssh/SshClientSettings.cs
```diff
@@ -536,6 +538,72 @@ public int DefaultWindowSize
...
+    /// </summary>
+    /// <remarks>
+    /// <para>Defaults to <see cref="ForwardMode.Off"/>. This can be overridden per remote process using <see cref="ExecuteOptions.ForwardX11"/>.</para>
+    /// <para>When set to <see cref="ForwardMode.Request"/>, X11 setup failures are logged and the remote process is started without X11 forwarding.</para>
+    /// <para>When set to <see cref="ForwardMode.Require"/>, X11 setup failures fail the operation.</para>
+    /// </remarks>
+    public ForwardMode ForwardX11 { get; set; } = ForwardMode.Off;
+
+    /// <summary>
+    /// Gets or sets whether remote X11 clients have full access to the local X11 display.
```
tmds: "Gets or sets whether to give remote X11 clients full access to the local X11 display."

src/Tmds.Ssh/SshConfigOption.cs
```diff
@@ -167,5 +167,25 @@ public enum SshConfigOption
     /// <summary>
     /// Forward the SSH agent to the server.
     /// </summary>
-    ForwardAgent
+    ForwardAgent,
+
+    /// <summary>
+    /// Whether X11 connections are forwarded for remote processes.
+    /// </summary>
```
tmds, line 174: "Whether to forward X11 connections"

src/Tmds.Ssh/SshConfigOption.cs
```diff
@@ -167,5 +167,25 @@ public enum SshConfigOption
...
+    ForwardX11,
+
+    /// <summary>
+    /// Whether remote X11 clients have full access to the local X11 display.
```
tmds: Whether to give remote X11 clients full access to the local X11 display.

src/Tmds.Ssh/SshConfigOption.cs
```diff
@@ -167,5 +167,25 @@ public enum SshConfigOption
...
+    /// </summary>
+    ForwardX11Trusted,
+
+    /// <summary>
+    /// Timeout for untrusted X11 forwarding. X11 connections received after this time are refused.
```
tmds: Timeout for untrusted X11 forwarding.

src/Tmds.Ssh/SshSession.cs
```diff
@@ -35,6 +35,8 @@ sealed partial class SshSession
     private int _keepAliveCount;
     private Dictionary<ListenAddress, RemoteListenerInfo>? _remoteListeners;
     private string? _forwardAgentAddress;  // Address of the agent to forward, or 'null' when agent forwarding is disabled.
+    private X11Forwarding? _x11Forwarding;
+    private static readonly ExecuteOptions NoX11ForwardingOptions = new() { ForwardX11 = ForwardMode.Off }; // Used for the sftp subsystem, which doesn't need X11 forwarding.
```
tmds: Let's name this SftpExecuteOptions.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+
+namespace Tmds.Ssh;
+
+// Reads X11 authorization data from Xauthority files, and generates untrusted authorization data using xauth.
+static class XAuthority
+{
+    public const string MitMagicCookie = "MIT-MAGIC-COOKIE-1";
```
tmds: Let's move the value to AuthenticationProtocol, and use X11Forwarding.AuthenticationProtocol in this file.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    // Address families from Xauth.h.
+    internal const ushort FamilyInternet = 0;
+    internal const ushort FamilyInternet6 = 6;
+    internal const ushort FamilyLocal = 256;
+    internal const ushort FamilyWild = 65535;
```
tmds: Can you add a comment `// internal for testing`, for APIs which are internal for tests?

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    // Like OpenSSH, the xauth timeout is extended so the cookie outlives ForwardX11Timeout.
+    private const int UntrustedTimeoutSlackSeconds = 60;
```
tmds: Let's move this private field to the top of the file.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+
+    private static readonly TimeSpan XAuthTimeout = TimeSpan.FromSeconds(30);
```
tmds: Let's make this one a property so it doesn't need storage.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+    internal readonly record struct Entry(ushort Family, byte[] Address, string Number, string Name, byte[] Data);
+
+    public static string GetDefaultFilePath()
+    {
+        string? path = Environment.GetEnvironmentVariable("XAUTHORITY");
+        return string.IsNullOrEmpty(path) ? Path.Combine(SshClientSettings.Home, ".Xauthority") : path;
+    }
+
+    // Returns the MIT-MAGIC-COOKIE-1 data for the display, or 'null' when there is none.
+    public static async Task<byte[]?> FindCookieAsync(string filePath, X11Display display, CancellationToken cancellationToken)
```
tmds: Let's move this method and GetAddressesAsync into X11Forwarding.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+            return null;
+        }
+
+        List<(ushort Family, byte[] Address)> addresses = await GetAddressesAsync(display, cancellationToken).ConfigureAwait(false);
+        return FindCookie(ParseEntries(content), display.DisplayNumber, addresses);
+    }
+
+    internal static byte[]? FindCookie(IEnumerable<Entry> entries, int displayNumber, List<(ushort Family, byte[] Address)> addresses)
+    {
+        string number = displayNumber.ToString(CultureInfo.InvariantCulture);
```
tmds: Let's parse entry.Number instead.

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+               TryReadBytes(ref content, out ReadOnlySpan<byte> name) &&
+               TryReadBytes(ref content, out ReadOnlySpan<byte> data))
+        {
+            entries.Add(new Entry(family, address.ToArray(), Encoding.ASCII.GetString(number), Encoding.ASCII.GetString(name), data.ToArray()));
+        }
+        return entries;
+
+        static bool TryReadUInt16(ref ReadOnlySpan<byte> content, out ushort value)
+        {
+            if (content.Length < 2)
```
tmds: Can we call BinaryPrimitives.TryReadUInt16BigEndian?

src/Tmds.Ssh/XAuthority.cs
```diff
@@ -0,0 +1,241 @@
...
+using System.Diagnostics;
+using System.Globalization;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+
+namespace Tmds.Ssh;
+
+// Reads X11 authorization data from Xauthority files, and generates untrusted authorization data using xauth.
+static class XAuthority
```
tmds: Let's move the members of this class into X11Forwarding. I don't think it is a clear split what belongs where so I prefer to just put it all together.

src/Tmds.Ssh/X11Display.cs
```diff
@@ -0,0 +1,134 @@
...
+                host = host[1..^1];
+            }
+            hostName = host.ToString();
+        }
+
+        display = new X11Display(name, hostName, socketPath, displayNumber, screenNumber);
+        return true;
+    }
+
+    public async Task<Stream> ConnectAsync(CancellationToken cancellationToken)
```
tmds: Let's move the ConnectAsync method into the X11Forwarding so this type becomes a pure data type.

src/Tmds.Ssh/X11Display.cs
```diff
@@ -0,0 +1,134 @@
...
+// See file LICENSE for full license details.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Net.Sockets;
+
+namespace Tmds.Ssh;
+
+// An X11 display as specified by the DISPLAY environment variable: [host]:displaynumber[.screennumber].
+sealed class X11Display
```
tmds: Let's make this a type that is nested in X11Forwarding. You can keep it in its own file if you like, `X11Forwarding.X11Display.cs`.

src/Tmds.Ssh/X11Display.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    public int ScreenNumber { get; }
+
+    // The display name to pass to xauth.
+    public string XAuthDisplayName
+    {
+        get
+        {
+            // Like OpenSSH: xauth doesn't find entries for 'localhost:N' and the launchd socket path.
+            if (Host == "localhost")
```
tmds: Should we do an invariant case-insensitive compare?

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,289 @@
...
+            displayName = Environment.GetEnvironmentVariable("DISPLAY");
+        }
+        if (string.IsNullOrEmpty(displayName))
+        {
+            throw new SshOperationException("X11 forwarding requires a display: DISPLAY is not set.");
+        }
+        if (!X11Display.TryParse(displayName, out X11Display? display))
+        {
+            throw new SshOperationException($"X11 forwarding failed: can not parse display '{displayName}'.");
+        }
```
tmds, line 74: Can we create the X11Display once in the constructor?

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,289 @@
...
+sealed class X11Forwarding
+{
+    public const string AuthenticationProtocol = XAuthority.MitMagicCookie;
+
+    private const int SetupHeaderLength = 12;
+    private const int FakeCookieLength = 16;
+
+    private static readonly byte[] AuthenticationProtocolBytes = Encoding.ASCII.GetBytes(AuthenticationProtocol);
+
+    internal sealed class Target
```
tmds: Let's rename Target to AuthBinding (and similar in fields, methods, properties).

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,289 @@
...
+        get
+        {
+            lock (_gate)
+            {
+                return _targets.Count > 0;
+            }
+        }
+    }
+
+    public async Task<Target> GetTargetAsync(SshClientSettings settings, CancellationToken cancellationToken)
```
tmds: "GetTargetAsync" -> GetOrCreateAuthBinding, and let's pass in the settings properties this uses instead of all the settings.

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,289 @@
...
+        {
+            _targets.Add(target);
+        }
+        return target;
+    }
+
+    public void HandleConnection(SshDataStream channelStream, string originatorAddress, uint originatorPort)
+        => _ = ForwardConnectionAsync(channelStream, $"{originatorAddress}:{originatorPort}");
+
+    private async Task ForwardConnectionAsync(SshDataStream channelStream, string sourceAddress)
```
tmds: Let's accept the CancellationToken as an argument of this method (instead of having a field dor it).

src/Tmds.Ssh/SshSession.cs
```diff
@@ -1135,11 +1165,43 @@ public async Task<SshDataStream> OpenUnixConnectionChannelAsync(string path, int
...
+        }
+
+        X11Forwarding.Target target;
+        try
+        {
+            target = await x11Forwarding.GetTargetAsync(_settings, cancellationToken).ConfigureAwait(false);
+        }
+        catch (SshOperationException ex) when (!isRequired)
+        {
+            Logger.X11ForwardingSetupFailed(ex);
```
tmds, line 1195: When isRequired, we should throw SshChannelException (since there is no usable channel).

src/Tmds.Ssh/X11Forwarding.X11Display.cs
```diff
@@ -0,0 +1,106 @@
...
+                }
+                if (SocketPath is not null)
+                {
+                    return $":{DisplayAndScreen}";
+                }
+                return Name;
+            }
+        }
+
+        private string DisplayAndScreen => Name.Substring(Name.LastIndexOf(':') + 1);
```
tmds:
```suggestion
private string DisplayAndScreen => ScreenNumber == 0
    ? DisplayNumber.ToString()
    : $"{DisplayNumber}.{ScreenNumber}";
```

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,541 @@
...
+        if (!X11Display.TryParse(displayName, out X11Display? display))
+        {
+            throw new SshOperationException($"X11 forwarding failed: can not parse display '{displayName}'.");
+        }
+
+        Display = display;
+        _logger = logger;
+    }
+
+    public X11Display Display { get; }
```
tmds: Does something need this property?

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,541 @@
...
+        {
+            displayName = Environment.GetEnvironmentVariable("DISPLAY");
+        }
+        if (string.IsNullOrEmpty(displayName))
+        {
+            throw new SshOperationException("X11 forwarding requires a display: DISPLAY is not set.");
+        }
+        if (!X11Display.TryParse(displayName, out X11Display? display))
+        {
+            throw new SshOperationException($"X11 forwarding failed: can not parse display '{displayName}'.");
```
tmds: Let's not throw. Store the Display as a nullable field. And add a property named HasDisplay that checks for null.

src/Tmds.Ssh/X11Forwarding.cs
```diff
@@ -0,0 +1,541 @@
...
+    public async Task<AuthBinding> GetOrCreateAuthBindingAsync(bool isTrusted, string? xauthorityFilePath, string xauthLocation, TimeSpan timeout, CancellationToken cancellationToken)
+    {
+        // Serialize so concurrent calls share a single binding instead of each generating authentication data.
+        await _authBindingSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
+        try
+        {
+            lock (_gate)
+            {
+                // Connections for expired bindings are refused, so they no longer need to be kept.
+                _authBindings.RemoveAll(authBinding => authBinding.IsExpired);
```
tmds: Let's remove _authBindings and replaces it by two nullable fields _trustedBinding and _untrustedBinding.

src/Tmds.Ssh/X11Forwarding.cs
joesdu: Only for the screen number in the x11-req. I replaced it with a `ScreenNumber` property and made `Display` private.

src/Tmds.Ssh/X11Forwarding.cs
joesdu: Done. `SshSession` now checks `HasDisplay` and handles it the same way as the other setup failures.

src/Tmds.Ssh/X11Forwarding.cs
joesdu: Done. An expired binding now just gets replaced. I also dropped the `_gate` lock since it's only two reference fields now and writes are already serialized by the semaphore. Let me know if you'd rather keep the lock.

PR #515, opened 2026-09-15
ssh example: demonstrate Wayland forwarding using waypipe

joesdu:
## Summary

Follow-up to https://github.com/tmds/Tmds.Ssh/pull/514#issuecomment-5675611238: rather than making Wayland forwarding a library feature, this demonstrates it in the `ssh` example application, in a form that can be copied into an application.

Wayland forwarding is done with [waypipe](https://gitlab.freedesktop.org/mstoeckl/waypipe), like `waypipe ssh` does:

```
ssh --forward-wayland myhost weston-terminal
```

- a local `waypipe client` listens on a Unix socket (in `XDG_RUNTIME_DIR`) and connects to the local Wayland compositor,
- the remote Unix socket `/tmp/waypipe-server-<id>.sock` is forwarded to it using `SshClient.StartRemoteForwardAsync`,
- the command is run by `waypipe server`, which sets `WAYLAND_DISPLAY` for it.

Everything lives in [`src/ssh/WaypipeForward.cs`](https://github.com/tmds/Tmds.Ssh/blob/main/src/ssh/WaypipeForward.cs) and only uses the public Tmds.Ssh API, so there are **no library changes and no public API changes** in this PR.

## Details

- Supported for commands and for a shell (`--login-shell server` runs the user's login shell).
- The waypipe invocation runs under `/bin/sh` with the user command passed as a positional argument, so it works with non-POSIX login shells and the command is still run by the user's `$SHELL`.
- When the server doesn't set `XDG_RUNTIME_DIR`, the display is placed at `/tmp/wayland-<id>`.
- Compression is passed explicitly to both ends (`--compress=lz4` by default): the defaults differ between waypipe versions (0.8 server: none, 0.9 client: lz4) and a mismatch makes the client reject the connection.
- The local waypipe process, the remote forward and the local socket are cleaned up when the forward is disposed (the remote socket is removed by `--unlink-socket`).
- waypipe output is logged at debug level (`-vv`) and is included in the exception when the local waypipe fails to start.

## Requirements

- waypipe on the client and on the server, and a local Wayland compositor.
- A server that allows Unix socket (`streamlocal`) remote forwarding (`AllowStreamLocalForwarding`, enabled by default in OpenSSH) and runs commands with a POSIX `/bin/sh`.
- waypipe runs on Linux/BSD, so on Windows the client has to run inside WSL (WSLg provides the compositor).

## Testing

- `dotnet build src/ssh/ssh.csproj`: builds without new warnings, and `--help` lists the new option.
- The forwarding logic is the one from #514, which was verified end-to-end against a Debian 13 server (OpenSSH 10.0p2, waypipe 0.9.2) and a waypipe 0.8.6 server, running `wayland-info` and `weston-simple-shm` against the WSLg compositor, with no sockets or waypipe processes left behind. Here it is ported to the public API; I will re-run that end-to-end check on this branch and report back.

Happy to drop the `--forward-wayland` option and keep only the file, or to move the explanation elsewhere, whichever you prefer.


src/ssh/WaypipeForward.cs
```diff
@@ -0,0 +1,278 @@
...
+        }
+        sb.Append(" --socket ");
+        sb.Append(QuoteArgument(RemoteSocketPath));
+        // waypipe places a display name in XDG_RUNTIME_DIR. When the server doesn't set XDG_RUNTIME_DIR (e.g. no login session manager),
+        // use an absolute path instead. RemoteDisplay only contains [a-z0-9-] so it doesn't need to be quoted.
+        sb.Append(" --display \"${XDG_RUNTIME_DIR:+");
+        sb.Append(RemoteDisplay);
+        sb.Append("}${XDG_RUNTIME_DIR:-/tmp/");
+        sb.Append(RemoteDisplay);
+        sb.Append("}\"");
```
tmds: When `XDG_RUNTIME_DIR` is set, the expansions concatenate (for example: `wayland-xxx + /run/user/1000 = wayland-xxx/run/user/1000`).

```suggestion
        sb.Append(" --display \"$([ -n \"$XDG_RUNTIME_DIR\" ] && printf '%s' '");
        sb.Append(RemoteDisplay);
        sb.Append("' || printf '%s' '/tmp/");
        sb.Append(RemoteDisplay);
        sb.Append("')\"");
```

tmds:
@joesdu thanks for extending the application to demonstrate how to do this! It's very nice to run this and see it in action.

PR #518, opened 2026-09-18
Expose SSH PTY pixel dimensions without changing existing callers.

buchmiet:
## Summary

RFC 4254 `pty-req` and `window-change` already carry pixel width/height as `uint32`. Tmds.Ssh serialized those fields as unconditional zeros, so a frontend that knows the real geometry could not pass it through.

This change:

- adds `ExecuteOptions.TerminalWidthPixels` / `TerminalHeightPixels` (default `0` = unspecified; `ArgumentOutOfRangeException` on negative values);
- adds `RemoteProcess.SetTerminalSize(width, height, widthPixels, heightPixels)`; the existing two-argument overload still sends `0, 0`;
- threads the pixel fields through `pty-req` at session open and `window-change` on resize;
- rejects negative pixels on the four-argument `SetTerminalSize` **before** they are cast to `uint32` (`WriteUInt32((uint)value)`), so `-1` cannot become `4294967295` on the wire.

Existing callers keep the same source and the same on-the-wire zeros.

## Test plan

- [x] `TerminalPixelDimensionTests` + `PublicApiTest` on net10.0 Debug (10 passed): defaults, zero accepted, ExecuteOptions negatives, pty-req/window-change payload, `SetTerminalSize` negative width/height pixels
- [ ] Upstream CI across net8.0–net11.0
- [ ] Existing two-argument `SetTerminalSize` still compiles and sends unspecified pixels (`0, 0`)

Made with [Cursor](https://cursor.com)


buchmiet:
Opened in error; closing.

